PR #9, opened 2024-07-11
Fast and correct BPE algorithms

aneubeck:


crates/bpe/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Novel Algorithm
+
+At a first glance, it seems impossible to achieve `O(n)` complexity while preserving the encoding output of the original BPE algorithm, since the original BPE algorithm needs to first scan the full input before it can make any encoding decision.
+For instance, the sequence `abab` would be encoded as `ab ab` when the dictionary contains the tokens `a b ab ba bc abc babc ababc` ordered by frequency. But appending a single character `ababc` would result in a pretty different tokenization: `ab a bc`. So without looking ahead it seems impossible to properly tokenize the text.
+
+The solution is to track ALL encodings for all text prefixes. For our example `ababc` we would get:
+- `a` ------> `a`
+- `ab` -----> `ab`
+- `aba` ----> `ab a`
```
rewinfrey, line 101: **Question:** Why is the resulting encoding `ab a` rather than `a ba`? Is this the consequence of

> Then, one scans over the preliminary encoding and determines the smallest token id by which any pair of neighboring token ids could be replaced.
The left most of them is replaced with that token id.

I guess why is the single `a` token not considered the smallest token id in this case, given its position in the dictionary implies highest frequency?

aneubeck: If you want to encode `aba`, then you start with the byte tokens `a` `b` `a`.
Next, you look for potential merges. in this example you can choose between `ab` and `ba`.
Since `ab` is the smaller one of the two you replace that one first.
I.e. you get now `ab` `a`.
But at this point not possible merge operation exists anymore (`ba` is no longer possible)

rewinfrey: > Next, you look for potential merges. in this example you can choose between ab and ba.
Since ab is the smaller one of the two you replace that one first.

Thanks for clarifying, I understand now. That merge operation is producing the smallest token id for a sequence of individual byte tokens or already paired tokens starting with the left most element working towards the end of the sequence.

hendrikvanantwerpen (approved):
Generally looks good. The various tests are useful.

I'd need to spend quite a bit more time to understand all the details, but the code structure corresponds to the program set out in the readme. Together with the tests, it looks good to me.

PR #16, opened 2024-09-24
Add support for o200k tokenization

hendrikvanantwerpen:


crates/bpe/src/byte_pair_encoding.rs
```diff
@@ -192,7 +205,9 @@ impl BytePairEncoding {
         let mut token_starts = vec![0];
         let mut bytes_hash_to_token = FnvHashMap::default();
         for (i, token) in iter.enumerate() {
-            bytes_hash_to_token.insert(hash_bytes(&token), i as u32);
+            if let Some(j) = bytes_hash_to_token.insert(hash_bytes(&token), i as u32) {
+                eprintln!("collision: ({i}, {j})");
+            }
```
aneubeck, line 210: should the function return an error instead when this happens?

crates/bpe/src/byte_pair_encoding.rs
```diff
@@ -541,19 +554,73 @@ mod tests {
...
+    const BPE_CL100K_LEN: usize = 100256;
+    const BPE_O200K_LEN: usize = 199998;
+
+    /// Use this to find a hashing factor for [`hash_bytes`] that prevents collisions.
+    /// 1. Ensure all supported tokenizers are in the list.
+    /// 2. Update the hash factor in [`hash_bytes`].
+    /// 3. Run [`update_token_dicts`] tests below to update data files.
+    #[test]
+    #[ignore = "run manually to find a suitable hash factor"]
+    fn find_hash_factor() {
```
aneubeck, line 580: :)

crates/bpe/src/byte_pair_encoding.rs
```diff
@@ -541,19 +554,73 @@ mod tests {
...
+                println!("hash factor: {factor}");
+                return;
+            }
+        }
+    }
 
-    // TODO: Move the generation of the dictionary into some build procedure?
     #[test]
-    fn test_serialize() {
+    #[ignore = "run manually to update data files"]
```
aneubeck, line 600: in principle, we could let this test run normally, since it will fix the broken data file (and one will see in the diff that something has changed)

aneubeck (approved):
What are the benchmark numbers for o200k?
Did things get slower?

crates/bpe/src/byte_pair_encoding.rs
hendrikvanantwerpen: Yes, let's do that.

crates/bpe/src/byte_pair_encoding.rs
hendrikvanantwerpen: If this happens, the assertion below the loop will fail. I wonder if we want to provide a way to specify the factor as part of the API, if anyone wants to construct a BPE from their own dictionary. The hard-coded constant might make it harder to reuse this if users bring their own tokens. But I'll save that for a follow-up PR.

hendrikvanantwerpen:
> What are the benchmark numbers for o200k?
> Did things get slower?

Performance characteristics look similar.

_cl100k_

image

_o200k_

image

PR #19, opened 2024-09-27
Expose hash factor in API

hendrikvanantwerpen:
Having a hardcoded hash factor may prevent users from using this crate with other token sets.

This PR:
- Adds an optional parameter to BPE construction methods that allows setting a custom hash factor.
- Exposes the `find_hash_factor` function in the API so it's easier for users to reuse it when working with other token sets.

## Tasks

- [X] Expose hash factor in API.
- [X] Make `find_hash_factor` easier to use if you don't have the crate source checked out.

### DIFF
--- a/crates/bpe/src/byte_pair_encoding.rs
+++ b/crates/bpe/src/byte_pair_encoding.rs
@@ -63,6 +63,8 @@ pub struct BytePairEncoding {
     /// But we don't have efficient access to it and therefore store it here again.
     /// If there is none, then the value is set to u32::MAX.
     next_prefix_match: Vec<u32>,
+    /// Hash factor used to prevent hash collisions.
+    hash_factor: u64,
 }
 
 fn serialize_daac<S: Serializer>(
@@ -156,11 +158,7 @@ fn token_bytes<'a>(all_tokens: &'a [u8], token_starts: &[u32], token_id: u32) ->
     &all_tokens[token_range(token_starts, token_id)]
 }
 
-fn hash_bytes(bytes: &[u8]) -> u32 {
-    hash_bytes_with_factor(bytes, 17846336922010275747)
-}
-
-fn hash_bytes_with_factor(bytes: &[u8], factor: u64) -> u32 {
+fn hash_bytes(bytes: &[u8], factor: u64) -> u32 {
     let mut hasher = FnvHasher::default();
     bytes.hash(&mut hasher);
     // Note: we save 1/3 of space for the hashmap by only using the most significant bits of the hash.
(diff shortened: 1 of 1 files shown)


crates/bpe/src/byte_pair_encoding.rs
```diff
@@ -156,25 +158,48 @@ fn token_bytes<'a>(all_tokens: &'a [u8], token_starts: &[u32], token_id: u32) ->
...
+/// when constructing a [`BytePairEncoding`] from those tokens.
+#[cfg(all(feature = "tiktoken-rs", feature = "rand"))]
+pub fn find_hash_factor_from_tiktoken(bpe: &tiktoken_rs::CoreBPE, len: usize) -> u64 {
+    find_hash_factor(|i| bpe._decode_native(&[i]), len)
+}
+
+/// Find a suitable hash factor for a set of given tokens that prevents collisions when
+/// constructing a [`BytePairEncoding`] from those tokens.
+#[cfg(feature = "rand")]
+pub fn find_hash_factor(tokens: impl Fn(usize) -> Vec<u8>, len: usize) -> u64 {
```
aneubeck, line 179: since this is used pretty rarely, I would take something like `&[&[u8]]` or `Iterator<Item = Vec<u8>>` or some combination as input.

crates/bpe/src/byte_pair_encoding.rs
```diff
@@ -156,25 +158,48 @@ fn token_bytes<'a>(all_tokens: &'a [u8], token_starts: &[u32], token_id: u32) ->
...
     bytes.hash(&mut hasher);
     // Note: we save 1/3 of space for the hashmap by only using the most significant bits of the hash.
     // To make them unique for the given tokens, we have to add unfortunately another multiplication.
     ((hasher.finish().wrapping_mul(factor)) >> 32) as u32
 }
 
+/// Find a suitable hash factor for the given tiktoken dictionary that prevents collisions
+/// when constructing a [`BytePairEncoding`] from those tokens.
+#[cfg(all(feature = "tiktoken-rs", feature = "rand"))]
+pub fn find_hash_factor_from_tiktoken(bpe: &tiktoken_rs::CoreBPE, len: usize) -> u64 {
```
aneubeck, line 172: nit: maybe `hash_factor_for_tiktoken`?

aneubeck (approved):

crates/bpe/src/byte_pair_encoding.rs
hendrikvanantwerpen: good point, I've aligned it with the parameters of the constructor

PR #20, opened 2024-10-01
Polish benchmark and include some figures in README

hendrikvanantwerpen:
Polishes the benchmark and adds some of the result figures in the README.

I considered adding the HuggingFace tokenizers to the benchmark as well, but they don't have cl100k and/or o200k readily available. I can figure out how to build a tokenizer from the tiktoken tokens. That would require computing some merge lists, if I understand it correctly. But I'm not sure it's worth the effort. E.g. [this](https://medium.com/@sweety.tripathi13/comparing-gpt-tokenizers-968b60f5a72b) suggests their tokenizer is quite a bit slower than tiktoken anyway.

[Rendered README](https://github.com/github/rust-gems/blob/update-benchmark/crates/bpe/README.md).


crates/bpe/README.md
```diff
@@ -193,4 +197,39 @@ We compared our implementations with the tiktoken implementation on a MacBook Pr
 As can be seen, our Backtracking implementation beats the TikToken Rust implementation by ~4x.
 And even the fully dynamic programming solution is faster with a more consistent runtime.
 The tuned heap implementation is still quite competitive to TikToken (especially for smaller inputs).
-If the requirement of correct BPE output can be relaxed, then the Greedy approach or the minimal encoding approach are the clear winners.
\ No newline at end of file
+If the requirement of correct BPE output can be relaxed, then the Greedy approach or the minimal encoding approach are the clear winners.
+
+### Counting results
+
+Results for counting o200k tokens for random 10000 byte slices. The setup time of the interval encoder is comparable to backtracking. After setup counting of slices of the original data are approximately constant time.
```
aneubeck, line 204: `10000`?
The graph mentions different sizes along the x-axis.
Can you describe the set up a bit more clearly? (Was the total text fixed sized and then we picked a random sub-slice and the sub-slice size varied?)

crates/bpe/README.md
```diff
@@ -193,4 +197,39 @@ We compared our implementations with the tiktoken implementation on a MacBook Pr
...
+
+![counting runtime comparison](./benches/result/counting-o200k.svg)
+
+### Encoding results
+
+Results for encoding o200k tokens for random 1000 bytes. The backtracking encoder consistently outperforms tiktoken by a constant factor.
+
+![encoding runtime comparison](./benches/result/encoding-o200k.svg)
```
aneubeck, line 212: The colors are pretty difficult to distinguish. Can we use differently dotted lines, etc. to distinguish them more?

crates/bpe/README.md
```diff
@@ -193,4 +197,39 @@ We compared our implementations with the tiktoken implementation on a MacBook Pr
...
\ No newline at end of file
+If the requirement of correct BPE output can be relaxed, then the Greedy approach or the minimal encoding approach are the clear winners.
+
+### Counting results
+
+Results for counting o200k tokens for random 10000 byte slices. The setup time of the interval encoder is comparable to backtracking. After setup counting of slices of the original data are approximately constant time.
+
+![counting runtime comparison](./benches/result/counting-o200k.svg)
+
+### Encoding results
```
aneubeck, line 208: maybe add a worst case example for tiktoken?
(some string without whitespaces)

crates/bpe/README.md
```diff
@@ -193,4 +197,39 @@ We compared our implementations with the tiktoken implementation on a MacBook Pr
...
+
+Results for encoding o200k tokens for random 1000 bytes. The backtracking encoder consistently outperforms tiktoken by a constant factor.
+
+![encoding runtime comparison](./benches/result/encoding-o200k.svg)
+
+### Incremental encoding results
```
aneubeck, line 214: Nit: maybe merge with the previous section?

The important point here is not so much the incremental encoding, but the reverse encoding aspect I think.
I guess it requires a bit of explanation...

Oh... did I check that it returns the same result for "random" input?
E.g. when we have all whitespace, then the reverse encoder must move the longest merged token correctly to the front.

crates/bpe/README.md
```diff
@@ -193,4 +197,39 @@ We compared our implementations with the tiktoken implementation on a MacBook Pr
 As can be seen, our Backtracking implementation beats the TikToken Rust implementation by ~4x.
 And even the fully dynamic programming solution is faster with a more consistent runtime.
 The tuned heap implementation is still quite competitive to TikToken (especially for smaller inputs).
-If the requirement of correct BPE output can be relaxed, then the Greedy approach or the minimal encoding approach are the clear winners.
\ No newline at end of file
+If the requirement of correct BPE output can be relaxed, then the Greedy approach or the minimal encoding approach are the clear winners.
+
+### Counting results
```
aneubeck, line 202: Maybe mention that we are not using the original tiktoken implementation, but some fork which provides a wrapper... (maybe it is easy to migrate to the original implementation now?)

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
-| Heap         | 1900 µs  | ✔          |
-
-As can be seen, our Backtracking implementation beats the TikToken Rust implementation by ~4x.
-And even the fully dynamic programming solution is faster with a more consistent runtime.
-The tuned heap implementation is still quite competitive to TikToken (especially for smaller inputs).
-If the requirement of correct BPE output can be relaxed, then the Greedy approach or the minimal encoding approach are the clear winners.
+We ran several benchmarks to compare performance between different encoders and with the tiktoken library:
 
-### Counting results
+- The first measuers encoding runtime for our different encoders and the tiktoken Rust implementation.
```
aneubeck, line 188:
```suggestion
- The first measures encoding runtime for our different encoders and the tiktoken Rust implementation.
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
+  This shows a ~3.5x performance increase for our fastest correct encoder comapred to the tiktoken library.
```
aneubeck, line 189:
```suggestion
  This shows a ~3.5x performance improvement for our fastest correct encoder compared to the tiktoken library.
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
 
-Results for counting o200k tokens for random 10000 byte slices. The setup time of the interval encoder is comparable to backtracking. After setup counting of slices of the original data are approximately constant time.
+- The second measures incremental encoding runtime, where the text is built up byte-by-byte.
+  This mode is not available in tiktoken, which only supports counting/encoding a complete text.
 
-![counting runtime comparison](./benches/result/counting-o200k.svg)
+- The third measures interval counting runtime, where the token count for slices of an original text are determined.
```
aneubeck, line 194:
```suggestion
- The third measures interval counting runtime, where tokens of sub-slices of a fixed text are counted.
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
+  After the initial tokenization of the text, token counting for slices is typically constant time.
+  This mode is not available in tiktoken, which only supports counting/encoding a complete text.
+
+All benchmarks were run on a MacBook Pro M1.
```
aneubeck, line 198:
```suggestion
All benchmarks were run single-threaded on a MacBook Pro M1.
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
+
+### Encoding
+
+Encoding is computing the tokens for a given text.
+This benchmark uses several encoders:
```
aneubeck, line 203:
```suggestion
This benchmark compares several encoders:
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
 
-### Encoding results
+- The backtracking encoder uses a backtracking algorithm based on a string matching automaton.
```
aneubeck, line 205:
```suggestion
- The backtracking encoder uses the backtracking algorithm with memorisation based on top of a string matching automaton.
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
+- The heap encoder uses a priority heap to implement the traditional BPE algorithm.
```
aneubeck, line 206:
```suggestion
- The heap encoder uses a priority heap and a bitmask to represent token positions to implement the traditional BPE algorithm.
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
+- The table encoder uses a dynamic programming algorithm.
```
aneubeck, line 207:
```suggestion
- The table encoder implements the raw dynamic programming algorithm proposed above.
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
 
-Results for encoding o200k tokens for random 1000 bytes. The backtracking encoder consistently outperforms tiktoken by a constant factor.
+Two additional encoders are included that are faster but do not always give exact results:
```
aneubeck, line 209:
```suggestion
Two additional encoders are included that are faster but deviate from the original BPE encoding strategy:
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
+
+- The greedy encoder uses a left-to-right greedy algorithm.
```
aneubeck, line 211:
```suggestion
- The greedy encoder picks the left-longest token.
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
+- The minimal encoder computes an encoding with the minimal number of tokens.
+
+The benchmark measured the runtime of encoding of slices of lengths 10, 100, 1000, and 1000 from a random 20000 token original text using the o200k token set.
+(All encodings were computed from scratch for each slice.)
+
+The graph below shows encoding runtime vs slice length.
+All encoders show similar runtime increases with increasing slice length.
```
aneubeck, line 218:
```suggestion
All encoders show the expected linear runtime complexity.
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
+The backtracking encoder, the fastest encoder that still returns correct results, shows a performance gain of approximately 3.5x compared to tiktoken.
+The fully dynamic programming solution and the heap implementation are still quite competitive to TikToken (especially for smaller inputs).
+If the requirement of correct BPE output can be relaxed, then the Greedy approach or the minimal encoding approach are the clear winners.
 
 ![encoding runtime comparison](./benches/result/encoding-o200k.svg)
 
-### Incremental encoding results
+### Incremental encoding
+
+Incremental encoding tokenizes a text to which bytes are appended.
```
aneubeck, line 227:
```suggestion
Incremental encoding tokenizes a text while appending bytes. This type of algorithm is interesting for use cases where a certain token budget must not be exceeded.
```

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
+This benchmark uses two encoders:
 
-Results for incrementally encoding o200k tokens by appending 10000 random bytes. The appending encoder is slower by a constant factor but overall has similar performance curve as the backtracking encoder encoding all data at once.
+- The backtracking encoder, which retokenizes the text froms cratch every time it changes.
```
aneubeck, line 230: mmm. this statement is not correct :)
Whenever it detects that it made a decision which is no longer compatible with new bytes, then it backtracks.
But it backtracks with memorisation which means it tries to compute only those states from the dynamic programming encoder which are actually necessary to find the correct solution...

crates/bpe/README.md
```diff
@@ -183,40 +183,82 @@ On average it is about ~4 faster, since the short-cuts usually pay off.
...
+
+Interval counting is counting the tokens for a slice of an original text.
+This benchmark uses two encoders:
+
+- The backtracking encoder encodes the slice from scratch.
+  This is similar to what one has to do with other libraries, like `tiktoken`.
+- The interval encoder encodes the original text once and reuses that encoding to count tokens for intervals of the original text.
+  The initial encoding time for the interval encoder is comparable to that of the backtracking encoder.
+
+The benchmark measured the runtime of counting o200k tokens on slices of lengths 10, 100, 1000, and 1000 from a random 20000 token original text.
```
aneubeck, line 254:
```suggestion
The benchmark measured the runtime of counting o200k tokens on slices of lengths 10, 100, 1000, and 10000 from a random 20000 token original text.
```

aneubeck (approved):
Depending on what we will focus on in the blog post, we might need to add more numbers (like worst-case inputs for tiktoken)

crates/bpe/README.md
hendrikvanantwerpen: I've added a note about that in 537699192c9eff01f906e785f549f06eb88217af.

crates/bpe/benches/performance.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+fn encoding_benchmark(c: &mut Criterion) {
+    for (name, bpe, tiktoken) in TOKENIZERS.iter() {
+        let text = create_test_string(&bpe, 20000);
+        let input = text.as_bytes();
+
+        let mut group = c.benchmark_group(format!("encoding-{name}"));
+        group.plot_config(PlotConfiguration::default().summary_scale(AxisScale::Logarithmic));
```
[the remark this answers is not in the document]
hendrikvanantwerpen: Not sure what you mean here? Are you talking about this bit of source or the output plot?

crates/bpe/README.md
hendrikvanantwerpen: Not sure what you mean? The appending encoder doesn't do reverse encoding afaik. We also have the prepending one, although that one's not included in the benchmark now.

crates/bpe/README.md
hendrikvanantwerpen: I tried that. This was the same as the encoding benchmark but all inputs were taken from a random ascii string without whitespace. The factor increased a bit (close to 6x) but the curves seem fairly similar to the encoding results.

aneubeck: mmm. What I tested some time ago was a string which was the concatenation of all unicode characters. That input never finished with the tiktoken lib... I think the regex simply returned a super large sub-chunk on which the quadratic encoder was then running. This obviously will take forever...

crates/bpe/benches/performance.rs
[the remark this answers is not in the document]
aneubeck: the github UI was rendering this line very awkwardly. Now it seems to be OK again.

crates/bpe/README.md
hendrikvanantwerpen: Maybe the ascii is too simple then. I found a way to sample unicode. I'll try that and see if it makes a difference.

hendrikvanantwerpen: I could not replicate what you're describing using random Unicode strings. I'll leave this for now and maybe get back to it if we want to highlight this on the blog post.

aneubeck: well, I didn't use random unicode strings... 🤷

PR #21, opened 2024-10-02
Append from interval encoding

hendrikvanantwerpen:


hendrikvanantwerpen:
Closing because this is not so straight-forward as I thought.

PR #24, opened 2024-10-04
Generate serialized data in build script

hendrikvanantwerpen:
Instead of including the serialized data in the crate, @aneubeck suggested that we generate it at build time.

This PR introduces a new `bpe-openai` crate that generates serialized data as part of its build. This cannot be part of `bpe` directly, because that would require `bpe` to be available when `bpe` is built. The `bpe-openai` crate exposed public functions to get static BPE instances loaded from the serialized data generated at build time. As a convenience, it re-exports the `bpe` crate, so depending on this crate is enough to be able to use these tokenizers fully.

Serializing the data is a hit on compile time. But I think this shouldn't happen too often with the `rerun-if-change:build.rs` directive.


crates/bpe-openai/README.md
```diff
@@ -0,0 +1,42 @@
+# OpenAI Byte Pair Encoders
+
+Fast tokenizers for OpenAI token sets based on the [bpe](https://crates.io/crates/bpe) crate.
+Serialized BPE instances are generated during build and lazily loaded at runtime as static values.
+The overhead of loading the tokenizers is small because it happens only once per proces and only requires deserialization (as opposed to actually building the internal datastructures).
```
tclem, line 5:
```suggestion
The overhead of loading the tokenizers is small because it happens only once per process and only requires deserialization (as opposed to actually building the internal data structures).
```

PR #27, opened 2024-10-09
Reorganize benchmark to include fairer comparisons

hendrikvanantwerpen:
This reorganizes the benchmark to make the comparisons more fair. Either all tokenizers in a test use pre-tokenization, or none do.

Code changes:

- The `bpe-openai` crate now includes a `Tokenizer` type that has a similar interface as other tokenization crates. It implements pre-tokenization and thus produces exactly the same results as tiktoken.
- The benchmarks are now their own crate, so I could depend on the `bpe-openai` crate without introducing a cyclic dependency.


crates/bpe-openai/src/lib.rs
```diff
@@ -1,42 +1,103 @@
...
+        "\\s+(?!\\S)",
+        "\\s+",
+    ].join("|");
+    Tokenizer::new(bpe, Some(&pat)).expect("valid regex")
 });
 
 pub use bpe::*;
 
-pub fn r50k() -> &'static BytePairEncoding {
+pub struct Tokenizer {
```
hendrikvanantwerpen, line 45: We could move this into `bpe`. That would allow us to serialize the regex as well.

aneubeck, line 45: There must be some more documentation to this struct IMO.
E.g. that the functions on this struct produce the same encodings as the original algorithms (at least in theory :) ).
But it also gives access to the two underlying data structures (the bpe encoding and the regex) if needed.
The bpe encoding can be used to tokenize and count tokens in more complex scenarios, but it will NOT take the regex splitting into account and therefore produce a DIFFERENT encoding solution.

aneubeck: I would not move this into `bpe`, since the whole idea of the `bpe` algorithms is to NOT have a regex :)

crates/bpe-openai/src/lib.rs
```diff
@@ -1,42 +1,103 @@
...
+    /// The byte-pair encoding for this tokenizer.
+    pub bpe: BytePairEncoding,
+    /// The pattern regex used to split the input.
+    pub pat: Option<Regex>,
+}
+
+impl Tokenizer {
+    #[allow(clippy::result_large_err)]
+    pub fn new(bpe: BytePairEncoding, pat: Option<&str>) -> fancy_regex::Result<Self> {
```
aneubeck, line 54: Question: did you test different regex libraries? Is this the fastest?

crates/bpe/README.md
```diff
@@ -25,15 +24,17 @@ BPE counting is unfortunately non-monotonic, i.e. appending more text could resu
...
 2) Heap based. Set up a heap with the frequencies. This improves the linear search time to a logarithmic factor. If done properly, the overall complexity reduces now to `O(n log n)`.
-3) Split the input into sections of a maximum size first and then process each section individually. This shrinks in theory the complexity to `O(n)` if the section size is small enough. But it will in general produce now different results. In order to produce the "correct" encoding, one would need to choose split points at token boundaries. But without having the text encoded already, this is in general impossible.
+3) Split the input into sections of a maximum size first and then process each section individually. This shrinks in theory the complexity to `O(n)` if the section size is small enough. But it will in general produce now different results. In order to produce the "correct" encoding, one would need to choose split points at token boundaries. But without having the text encoded already, this is in general impossible. (Note that tiktoken as well as other tokenizers often split the input as part of pre-tokenization to improve model performance.)
```
aneubeck, line 37: > (Note that tiktoken as well as other tokenizers often split the input as part of pre-tokenization to improve model performance.)

Do you have a reference for this statement :)
Otherwise, I wouldn't claim that this was the reason why tiktoken did it...

crates/bpe/README.md
```diff
@@ -219,19 +221,18 @@ Two additional encoders are included that are faster but deviate from the origin
 The benchmark measured the runtime of encoding of slices of lengths 10, 100, 1000, and 10000 from a random 20000 token original text using the o200k token set.
 (All encodings were computed from scratch for each slice.)
 
+Be aware that this benchmark none of the tokenizers pre-tokenize the input.
```
aneubeck, line 224:
```suggestion
Be aware that in this benchmark none of our(?) tokenizers pre-tokenize the input.
```

crates/bpe/README.md
```diff
@@ -264,10 +265,47 @@ The graph below shows counting runtime vs slice length.
...
+### Comparison with other tokenizers
+
+We compared the encoding performance of our encoder with two popular implementations, tiktoken and Huggingface tokenizers.
+
+The benchmark measured the runtime of encoding of slices of lengths 10, 100, 1000, and 10000 from a random 20000 token original text using the o200k token set.
+(All encodings were computed from scratch for each slice.)
+
+In this benchmark all tokenizers pre-tokenize their input and produce the same tokens and decoded texts as the tiktoken tokenizer.
+An effect of pre-tokenization is that the inputs to the actual BPE logic are typically much smaller than the overall input size, especially for larger inputs.
+It is therefore difficult to judge the performance differences of the BPE logic fromt his benchmark.
```
aneubeck, line 279:
```suggestion
It is therefore difficult to judge the performance differences of the BPE logic from this benchmark.
```

crates/bpe-openai/src/lib.rs
hendrikvanantwerpen: I didn't, this is the same library tiktoken uses. The regex uses negative lookahead though, which isn't supported by many libraries.

hendrikvanantwerpen: Looks like someone has a [PR](https://github.com/openai/tiktoken/pull/331) on tiktoken to get rid of `fancy-regex`. But at the expense of pushing some of that logic into the code.

I wonder how complex the state machine for these regexes is. Perhaps not too complex if you can reuse regex logic for the character classes?

crates/bpe/README.md
hendrikvanantwerpen: I thought I had it from [here](https://huggingface.co/blog/catherinearnett/dangers-of-tokenizer-recycling#pre-tokenizer), but it doesn't actually say why tiktoken or any of the others use it. On the other hand, I haven't found a reference either suggesting that pre-tokenization was done to improve tokenization performance.

crates/bpe/README.md
```diff
@@ -264,10 +265,47 @@ The graph below shows counting runtime vs slice length.
...
+It does give a good indication of how the algorithms might perform in practice.
+
+The graph below shows encoding runtime vs slice length.
+All encoders (except the heap encoder) show the expected linear runtime complexity.
+The backtracking encoder, the fastest encoder that still returns correct results, shows a performance gain of approximately 3.5x compared to tiktoken.
+The fully dynamic programming solution and the heap implementation are still quite competitive to TikToken (especially for smaller inputs).
+If the requirement of correct BPE output can be relaxed, then the Greedy approach or the minimal encoding approach are the clear winners.
```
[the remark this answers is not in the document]
hendrikvanantwerpen: You're right, that shouldn't be there

crates/bpe/README.md
hendrikvanantwerpen: This is the closest I could find to anything on pre-tokenization: https://arxiv.org/abs/2402.01035. They study the effect of tokenization choices on model performance and note:

> Splitting sequences prevents BPE from merging certain tokens, for instance splitting on white spaces means that a token cannot span two space-separated words. It leads to shorter tokens and thus worse compression rates, but is generally done to improve downstream performance.

crates/bpe-openai/src/lib.rs
hendrikvanantwerpen: I was about to argue that this pre-tokenization is probably not going anywhere, since this seems to be done to improve model (quality) performance, and would require integrating these _more_.

However, when writing that down, I realized that all of that can probably be achieved by _wrapping_ the BPE parts and doing pre-tokenization logic around it.

So, I agree that we should keep it out.

aneubeck:
Another thought on pretokenization...
I would probably argue that using pretokenization should be used to train the BPE encoding, i.e. for building the dictionary.
But afterwards, I'm very doubtful that it is actually needed. It's probably done because nobody thought about doing dropping this step JUST for encoding 🤷 

Reasoning is: if you don't have tokens in the dictionary which cross certain character boundaries, then BPE won't generate those anyways. There might be a subtle difference in what BPE outputs compared to regex. But, it is just so much simpler to have only ONE algorithm defining your output than two nested ones... Essentially this crate proves that if you simplify your requirements you can actually improve performance further (and get some additional benefits).

crates/bpe-openai/src/lib.rs
```diff
@@ -42,6 +42,12 @@ static BPE_O200K: LazyLock<Tokenizer> = LazyLock::new(|| {
 
 pub use bpe::*;
 
+/// A byte-pair encoding tokenizer that supports a pre-tokenization regex.
+/// The direct methods on this type pre-tokenize the input text and should
+/// produce the same output as the tiktoken tokenizers. The type gives access
+/// to the regex and underlying bye-pair encoding if needed. Note that using
```
aneubeck, line 48:
```suggestion
/// to the regex and underlying byte-pair encoding if needed. Note that using
```

crates/bpe/README.md
```diff
@@ -30,11 +30,12 @@ The comparison with the Rust tiktoken implementation is more subtle, because pre
...
-3) Split the input into sections of a maximum size first and then process each section individually. This shrinks in theory the complexity to `O(n)` if the section size is small enough. But it will in general produce now different results. In order to produce the "correct" encoding, one would need to choose split points at token boundaries. But without having the text encoded already, this is in general impossible. (Note that tiktoken as well as other tokenizers often split the input as part of pre-tokenization to improve model performance.)
+
+Note that many tokenizers split the input into sections and then process each section individually. This shrinks in theory the complexity to `O(n)` if the section size is small enough. But it will in general produce now different results. In order to produce the "correct" encoding, one would need to choose split points at token boundaries. But without having the text encoded already, this is in general impossible. Input splitting may is therefore not a viable strategy for improving encoding performance.
```
aneubeck, line 38: didn't check whether we use the term `section` in other places. But I think `chunk` or `substring` would be the better term.

aneubeck (approved):

PR #32, opened 2024-10-17
Eliminate look-ahead in regex

hendrikvanantwerpen:
Eliminates look-ahead in the pre-tokenization regexes by supporting manually implemented trim functions in the pretokenizer.

The pretokenizer still accepts fancy regex, but since we don't use these features, it'll fall back to the regular regex crate.

## Tasks

- [x] Run benchmark, update figures and text.


hendrikvanantwerpen:
Closed in favor of #33.

PR #34, opened 2024-10-18
Move equivalence testing to bpe-openai

hendrikvanantwerpen:
Changes:

- Moved the tiktoken / bpe-openai equivalence tests into the bpe-openai crate, instead of the benchmarks crate.
  - The test fails for p50k. I only noticed now, since we were only testing cl100k & o200k before.
- Added the `create_test_string` function to `bpe` so it's easier to reuse and updated it so it can also generate strings with multi-byte characters now.


crates/bpe/src/byte_pair_encoding.rs
```diff
@@ -552,3 +552,74 @@ impl BytePairEncoding {
...
+        if let Some(i) = tokens.pop() {
+            bytes.truncate(bytes.len() - bpe.token_len(i as u32));
+            valid_bytes = bytes
+                .iter()
+                .rev()
+                .find_position(|b| is_char_boundary(**b))
+                .map_or(0, |(offset, _)| bytes.len() - (offset + 1));
+        }
+    }
+    // truncate to the know valid bytes
```
aneubeck, line 608:
```suggestion
    // truncate to the known valid bytes
```

crates/bpe/src/byte_pair_encoding.rs
```diff
@@ -552,3 +552,74 @@ impl BytePairEncoding {
...
+                .map_or(0, |(offset, _)| bytes.len() - (offset + 1));
+            assert!(last >= valid_bytes);
+            if std::str::from_utf8(&bytes[valid_bytes..last]).is_ok() {
+                tokens.push(i);
+                valid_bytes = last;
+                continue 'keep;
+            } else {
+                bytes.truncate(bytes.len() - bpe.token_len(i as u32));
+            }
+        }
```
aneubeck, line 597: First time actually looking at this code :)
I'm not so sure that this procedure has any reasonable chance of constructing valid utf8 from "broken" tokens.
Essentially there are VERY few of those tokens and the chance of picking the correct one is extremely low.
The original idea of picking tokens was that the chance of picking longer interesting token sequences is higher.
But, this doesn't really work with broken utf8 sequences so easily. You would have to store valid token pairs. But there are just too many pairs that this is computationally reasonable.

crates/bpe/tests/src/lib.rs
```diff
@@ -1,16 +1,3 @@
-use bpe::byte_pair_encoding::BytePairEncoding;
-use rand::{thread_rng, Rng};
-
-pub fn create_test_bytes(bpe: &BytePairEncoding, tokens: usize) -> Vec<u8> {
```
aneubeck, line 4: I "think" this function was testing more corner cases than the `create_test_string` variation 🤷

aneubeck (approved):

crates/bpe/src/byte_pair_encoding.rs
hendrikvanantwerpen: I see. I did a quick test and for both cl100k and o200k around 99% of the tokens are valid utf-8. o200k has more tokens containing multi-byte characters. For cl100k, almost 95% match the old condition `token.iter().all(|b| is_char_boundary(*b))`, but only 64% of o200k tokens does.

I guess I can simplify this function a bit by only testing for valid utf-8 tokens.

crates/bpe/tests/src/lib.rs
hendrikvanantwerpen: Probably true, yes. I'll see where I can reinstate that.

PR #35, opened 2024-10-21
Add count_till_limit method on Tokenizer

hendrikvanantwerpen:
Adds a `Tokenizer::count_till_limit` method that accounts for input splitting.

### DIFF
--- a/crates/bpe-openai/src/lib.rs
+++ b/crates/bpe-openai/src/lib.rs
@@ -86,6 +86,15 @@ impl Tokenizer {
             .sum()
     }
 
+    pub fn count_till_limit(&self, text: &str, token_limit: usize) -> Option<usize> {
+        self.split(text)
+            .try_fold(token_limit, |token_limit, piece| {
+                self.bpe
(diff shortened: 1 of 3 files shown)


crates/bpe-openai/src/lib.rs
```diff
@@ -86,6 +86,15 @@ impl Tokenizer {
             .sum()
     }
 
+    pub fn count_till_limit(&self, text: &str, token_limit: usize) -> Option<usize> {
```
aneubeck, line 89: some comment for the function?
It needs at least some explanation what the return value is...

aneubeck: (maybe also comments for the other pub functions)

hendrikvanantwerpen: Done in https://github.com/github/rust-gems/pull/35/commits/c28e428f76bf3605e08e8ae251f46d7c72f8a4d2

aneubeck (approved):

PR #38, opened 2024-11-01
WIP: Add `utf8-converter`

jorendorff:


crates/utf8-converter/README.md
```diff
@@ -0,0 +1,12 @@
+# UTF-8 Converter
+
+This crate converts string positions between Rust style (UTF-8 byte offsets) and styles used by other programming languages, as well as line numbers.
```
aneubeck, line 3: Can you expand this a bit further?
(or do you plan to write more in a subsequent PR)

E.g. that all (most) operations are constant after some linear preprocessing of the input text.
And that space consumption is x bytes per input byte.

crates/utf8-converter/src/bitrank.rs
```diff
@@ -0,0 +1,504 @@
...
+//! There is also an opportunistic `select` operation, but the general case has not been
+//! implemented.
+
+type Chunk = u128;
+
+// Static sizing of the various components of the data structure.
+const BITS_PER_BLOCK: usize = 16384;
+const BITS_PER_SUB_BLOCK: usize = 128;
+const SUB_BLOCKS_PER_BLOCK: usize = BITS_PER_BLOCK / BITS_PER_SUB_BLOCK;
+const BITS_PER_CHUNK: usize = 128;
```
aneubeck, line 13: note: I figured out a potentially faster solution here with a simple trick.
https://github.com/github/rust-gems/pull/37/files#diff-62430fbe79087bea140708a5d90481a016def9872a1180d9217781caa4c83499R113-R133
Ignore all the mapping stuff (it didn't actually reduce the number of page reads).
But you can store the sum up to the middle of a chunk and then either subtract the bits BEFORE the middle or add the bits AFTER the middle. That way, you can use u64 instead of u128 which in my benchmarks was faster.

Maybe something for a separate PR...

crates/utf8-converter/README.md
aneubeck: Also there was: https://github.com/aneubeck/algos/blob/main/crates/utf8-converter/README.md
I.e. Greg prepared the move ~1 year ago (forgot about that one). Not sure how much it differs from our internal version.

jorendorff: I'll steal Greg's text, and briefly say how to think about the performance.

I'll also link to [Tim's blog post](https://adaptivepatchwork.com/2023/07/10/utf-conversion/) about this.

jorendorff: I'm going to make those changes in github/blackbird#8230, though, and update this PR when that's all merged.

crates/utf8-converter/src/bitrank.rs
jorendorff: Yes, this is wild. Do you have those benchmarks lying around anywhere? I can throw them in while I'm here...

aneubeck: Sorry, didn't write any for the underlying bitrank stuff...

README.md
```diff
@@ -4,6 +4,7 @@ A collection of useful algorithms written in Rust. Currently contains:
 
 - [`geo_filters`](crates/geo_filters): probabilistic data structures that solve the [Distinct Count Problem](https://en.wikipedia.org/wiki/Count-distinct_problem) using geometric filters.
 - [`bpe`](crates/bpe): fast, correct, and novel algorithms for the [Byte Pair Encoding Algorithm](https://en.wikipedia.org/wiki/Large_language_model#BPE) which are particularly useful for chunking of documents.
+- [`utf8-converter`](crates/utf8-converter): converts string positions between bytes, chars, UTF-16 code units, and line numbers. Useful when sending string indices across language boundaries.
```
aneubeck, line 7: nit: this needs updating

crates/string-offsets/src/bitrank.rs
```diff
@@ -0,0 +1,499 @@
...
+        let mask = 1 << ((BITS_PER_SUB_BLOCK - 1) - bit_idx);
+        self.bits[chunk_idx] & mask != 0
+    }
+
+    /// The **total rank** of the block relative local index, and the index of the one
+    /// bit that establishes that rank (aka "select") **if** it occurs within that same
+    /// chunk, otherwise ['None'].  The assumption is that if you would have to look back
+    /// through previous chunks it would actually be cheaper to do a lookup in the original
+    /// data structure that the bit vector was created from.
+    fn rank_select(&self, local_idx: usize) -> (usize, Option<usize>) {
```
aneubeck, line 72: also here, if we don't need the `select` information, then just drop it...

crates/string-offsets/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "string-offsets"
+authors = ["The blackbird team <support@github.com>"]
+version = "0.1.0"
+edition = "2021"
+description = "Offset calculator to convert between byte, char, and line offsets in a string."
+repository = "https://github.com/github/rust-gems"
+license = "MIT"
+keywords = ["unicode", "string", "offsets", "positions", "interoperability"]
```
aneubeck, line 9: should we add utf8 utf16 runes characters, etc. to the list of keywords?

crates/string-offsets/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "string-offsets"
+authors = ["The blackbird team <support@github.com>"]
+version = "0.1.0"
+edition = "2021"
+description = "Offset calculator to convert between byte, char, and line offsets in a string."
```
aneubeck, line 6:
```suggestion
description = "Offset calculator to convert between byte (utf8 encoding), word (utf16 encoding), unicode characters, and line offsets in a string."
```

crates/string-offsets/README.md
```diff
@@ -0,0 +1,45 @@
+# string-offsets
+
+Offset calculator to convert between byte, char, and line offsets in a string.
```
aneubeck, line 3: see other comment

crates/string-offsets/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Rust strings are UTF-8, but JavaScript has UTF-16 strings, and in Python, strings are sequences of
+Unicode code points. It's therefore necessary to adjust string offsets when communicating across
+programming language boundaries. [`StringOffsets`] does these adjustments.
+
+Each `StringOffsets` value contains offset information for a single string. [Building the data
```
aneubeck, line 9:
```suggestion
Each `StringOffsets` instance contains offset information for a single string. [Building the data
```

crates/string-offsets/README.md
```diff
@@ -0,0 +1,45 @@
...
+structure](StringOffsets::new) takes O(n) time and memory, but then each conversion is fast.
```
aneubeck, line 10:
```suggestion
structure](StringOffsets::new) takes O(n) time and memory, but then most conversions are O(1).
```

aneubeck (approved):
nice!

crates/string-offsets/Cargo.toml
jorendorff: We only get five, but I like those, and probably we don't need "string" or "offsets". Will reshuffle.

jorendorff: (Also, I suspect these don't matter much.)

crates/string-offsets/Cargo.toml
jorendorff: I deleted some other stuff and ended up with:

> Converts string offsets between UTF-8 bytes, UTF-16 code units, Unicode code points, and lines.

PR #53, opened 2025-03-13
Add wasm package

mjbvz:
This lets us compile the lib into wasm and adds a npm wrapper package for it

I had to rearrange the code slightly to split out any methods using wasm incompatible types. Not sure if there's a better way to do this


crates/string-offsets/js/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "compile:nodejs": "wasm-pack build --target nodejs -d js/pkg/nodejs && node -e \"fs.unlinkSync('./pkg/nodejs/.gitignore')\"",
+    "compile": "npm run compile:web && npm run compile:bundler && npm run compile:nodejs",
+    "test": "jest"
+  },
+  "devDependencies": {
+    "@types/jest": "^29.5.14",
+    "jest": "^29.0.0",
+    "wasm-pack": "^0.13.1"
+  }
+}
```
tclem, line 34: **Nit:** Fix trailing newlines in a bunch of these files (makes diffs nicer going forward).

crates/string-offsets/js/test/test.js
```diff
@@ -0,0 +1,23 @@
...
+    });
+
+    test('Unicode text', () => {
+        const text = "☀️hello\n🗺️world";
+        const offsets = new StringOffsets(text);
+
+        expect(offsets.lines()).toBe(2);
+        // ☀️ is 6 UTF-8 bytes and 3 UTF-16 code units
+        expect(offsets.utf8_to_utf16(6)).toBe(2);
+        expect(offsets.utf8_to_utf16(0)).toBe(0);
```
tclem, line 21: **Question:** Do you want to use something like https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/js_name.html so that this doesn't feel like such a rust method?

tclem (approved):
Very cool, looks great to me. Having to move some methods to another impl block seems fine.

(Maybe as a follow up, we should add this to the Actions CI workflow)

crates/string-offsets/js/package-lock.json
```diff
@@ -0,0 +1,3870 @@
+{
+  "name": "string-offsets-dev",
```
aneubeck, line 2: Is there a reason why these have a `dev` suffix?

crates/string-offsets/CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
+# Contributing
+
+## Building the WASM/JS package
+
+The code for the wasm + js wrapper package is stored in the `js` directory. To build it:
+
+```sh
```
aneubeck, line 7: should there be a presubmit action which checks that the npm builds and the test passes?

crates/string-offsets/Cargo.toml
```diff
@@ -8,6 +8,13 @@ repository = "https://github.com/github/rust-gems"
 license = "MIT"
 keywords = ["unicode", "positions", "utf16", "characters", "lines"]
 categories = ["algorithms", "data-structures", "text-processing", "development-tools::ffi"]
+exclude = ["/js"]
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+wasm-bindgen = "0.2"
```
hendrikvanantwerpen, line 17: Question: do we want to have this as an unconditional dependency?

I spot checked a few other crates, and it seems more common to put it behind a feature flag or make it [conditional on a build target](https://github.com/seanmonstar/reqwest/blob/f63c6315911154de8b96b79012ce8e266e47dba6/Cargo.toml#L191-L196). However, doing it like this would make the inline attributes very anoying to write (they'd all have to be wrapped in conditionals as well), and a separate `wasm` module with the exported API that is conditionally included seems cleaner.

mjbvz: Ok good idea. I moved this to use a feature so that wasm-bindgen is now an optional dependency 

I'd love to split the was stuff out to its own module but wasn't sure how to do this without duplicating a whole bunch of code

tclem (approved):
LGTM

PR #58, opened 2025-03-25
Add bench and speed things up

aneubeck:
The compiler can generate faster code if the loop increment is not data dependent.
Most characters are anyways ascii.


crates/string-offsets/src/lib.rs
```diff
@@ -357,24 +362,21 @@ impl StringOffsets {
 
 fn new_converter(content: &[u8]) -> StringOffsets {
     let n = content.len();
-    let mut utf8_builder = BitRankBuilder::with_capacity(n);
-    let mut utf16_builder = BitRankBuilder::with_capacity(n);
-    let mut line_builder = BitRankBuilder::with_capacity(n);
+    let mut utf8_builder = BitRankBuilder::with_capacity(n + 1);
```
gorzell, line 365: **Question:** Why `+ 1`?

gorzell (approved):

crates/string-offsets/benchmarks/performance.rs
```diff
@@ -0,0 +1,32 @@
+use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion};
+use rand::{rng, Rng};
+use string_offsets::StringOffsets;
+
+fn construction_benchmark(c: &mut Criterion) {
+    // Generate random input string
+    let mut group = c.benchmark_group("construction");
+    for size in [1000, 10000, 100000] {
+        let mut rng = rng();
+        let random_input: String = (0..size).map(|_| rng.random::<u8>() as char).collect();
```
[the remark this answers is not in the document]
aneubeck: I think so. We could add another benchmark with only non ascii characters if we want.

crates/string-offsets/src/lib.rs
aneubeck: I moved all bits in this bitrank by 1. As a result, I need one more bit at the end...

PR #62, opened 2025-05-05
Add voyage3 model

aneubeck:
voyage3 requires unicode normalization for which I introduced basic support.
I also fixed the broken equivalence tests in the benchmarks folder.
It is important to note that huggingface uses some byte to char conversion, so that it can run the BPE algorithm on unicode strings. In the test, the pretokenizer were not properly disabled.

We could consider adding features for the different models.


CleanCut:
I pushed a commit which fixed a `cargo fmt` difference in Rust 1.86.

crates/bpe-openai/src/normalizer.rs
```diff
@@ -0,0 +1,56 @@
+use std::borrow::Cow;
+
+use unicode_normalization::UnicodeNormalization;
+
+/// Type which represents a normalized string.
+/// This is to avoid calling normalize multiple times of forgetting to call normalization!
```
jorendorff, line 6:
```suggestion
/// This is to avoid calling normalize multiple times or forgetting to call normalization!
```

crates/bpe-openai/src/normalizer.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /// This function is unsafe, since the caller must ensure that the correct normalization
+    /// was used. The normalization may vary by tokenizer. This mostly a backdoor which might
+    /// be handy for certain optimizations or for testing.
+    pub unsafe fn from_str(s: &'a str) -> NormalizedString<'a> {
+        // SAFETY: This is safe because we are creating a NormalizedString from a &str
+        // which is guaranteed to be valid UTF-8.
+        NormalizedString(Cow::Borrowed(s))
+    }
```
jorendorff, line 24:
```suggestion
        // SAFETY: This is safe if `s` is in fact correctly normalized already. The caller is
        // responsible for ensuring that.
        NormalizedString(Cow::Borrowed(s))
    }
```

crates/bpe-openai/src/normalizer.rs
```diff
@@ -0,0 +1,56 @@
+use std::borrow::Cow;
+
+use unicode_normalization::UnicodeNormalization;
+
+/// Type which represents a normalized string.
```
jorendorff, line 5:
```suggestion
/// Type which represents an NFC normalized string.
```

jorendorff (approved):
:+1:

jorendorff:
@aneubeck This is a breaking change, so it requires a version bump, right?

PR #69, opened 2025-07-18
Geofilter customize hasher

itsibitzi:


crates/geo_filters/src/diff_count.rs
```diff
@@ -92,7 +96,7 @@ impl<C: GeoConfig<Diff>> GeoDiffCount<'_, C> {
     ///
     /// Note: we need a peekable iterator, such that we can extract the most significant bits without
     /// having to construct another iterator with the remaining `BitChunk`s.
-    fn from_bit_chunks<I: Iterator<Item = BitChunk>>(config: C, chunks: I) -> Self {
+    fn from_bit_chunks<I: Iterator<Item = BitChunk>>(config: C, hash_builder: H, chunks: I) -> Self {
```
aneubeck, line 99: mmm. It's a bit annoying that the hash_builder has to be passed in everywhere now...
Would it be simpler if the `config` has the HashBuilder as member?

itsibitzi:
Update:

The new commit introduces a trait that library users must mark their `BuildHasher`s with in order to promise to us that the two `Hasher`s produced using the `default` method will produce the same output for the same input. 

The advantage of this is because we're not holding any state any more we can use an associated type on the config object and create a new `BuildHasher` whenever we need it. This makes the type parameters a lot cleaner and should not have performance impact since generally building a `BuildHasher` should be a no-op. I will verify this against benchmarks though.

The downside is from the library users perspective because it means they will need to create some newtype for the `BuildHasher` in order to get around the orphan problem (their `BuildHasher` is likely from a library, and our trait is also from a library). We could create either another crate to provide these newtypes or we could introduce feature flags to this crate for common hasher implementations.

crates/geo_filters/src/build_hasher.rs
```diff
@@ -0,0 +1,38 @@
...
+            "Hashers produced by GeoFilterBuildHasher do not produce the same output with the same input"
+        );
+    }
+}
+
+// Note that this `BuildHasher` has a consistent implementation of `Default`
+// but is NOT stable across releases of Rust. It is therefore dangerous
+// to use if you plan on serializing the geofilters and reusing them due
+// to the fact that you can serialize a filter made with one version and
+// deserialize with another version of the hasher factor.
```
aneubeck, line 34: nit: change into doc comments?

crates/geo_filters/src/diff_count/config.rs
```diff
@@ -17,7 +18,7 @@ use crate::Diff;
 //
 //     scripts/accuracy -n 10000 geo_diff/u16/b=7/bytes=112/msb={8,12,16,20}
 //
-pub type GeoDiffConfig7 = FixedConfig<Diff, u16, 7, 112, 12>;
+pub type GeoDiffConfig7 = FixedConfig<Diff, u16, 7, 112, 12, DefaultBuildHasher>;
```
aneubeck, line 21: Should `GeoDiffConfig7` have the BuildHasher as generic with `DefaultBuildHasher` set as the default?
(same for the others?)

itsibitzi: I was thinking about this, and presumably the accuracy figures from the comments are based on a combination of the config parameters and the specific hasher. E.g. previously I had `FnvBuildHasher` in here and the accuracy was way off when working against a series of incrementing numbers.

If we provide the pre-canned configs in order to meet a specified error - does it confuse things to allow people to swap out the hash function and potentially get a different error?

I suspect it is the case where any sufficiently good hash function/input combination should perform about the same?

crates/geo_filters/docs/choosing-a-hash-function.md
```diff
@@ -0,0 +1,113 @@
...
+`Default::default` will hash identical items to the same `u64` value across multiple calls
+to `BuildHasher::hash_one`.
+
+The following is an example of some incorrect code which produces nonsense results:
+
+```rust
+// Implement our marker trait for `RandomState`.
+// You should _NOT_ do this as `RandomState::default` does not produce
+// reproducible hashers.
+impl ReproducibleBuildHasher for RandomState {}
```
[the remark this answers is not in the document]
aneubeck: I "think" copilot is right...
Maybe doc generation is not correctly setup?

crates/geo_filters/src/diff_count/config.rs
aneubeck: When you use a bad hash function, then you get essentially no guarantees...
For all good ones, it should be about the same on average.
The amount of outliers might obviously differ 🤷

crates/geo_filters/src/lib.rs
```diff
@@ -29,13 +32,14 @@ pub struct Distinct {}
 impl Method for Distinct {}
 
 /// Trait for types solving the set cardinality estimation problem.
-pub trait Count<M: Method> {
+pub trait Count<M: Method, H: ReproducibleBuildHasher> {
```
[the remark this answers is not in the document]
itsibitzi: Good shout, I've made it an associated type :)

crates/geo_filters/src/diff_count/config.rs
itsibitzi: Ok, I thought so. 

I'll make the type alias generic and make another alias for using the default/unstable build hash so that the tests are less verbose.

itsibitzi: [Inference doesn't fall back to defaults](https://quinedot.github.io/rust-learning/misc-default-param.html) so this will introduce a few extra type annotations, mostly in the tests.

crates/geo_filters/src/lib.rs
```diff
@@ -29,13 +32,14 @@ pub struct Distinct {}
...
     /// Add the given hash to the set.
     fn push_hash(&mut self, hash: u64);
 
-    /// Add the hash of the given item, computed with the default hasher, to the set.
+    /// Add the hash of the given item, computed with the configured hasher, to the set.
     fn push<I: Hash>(&mut self, item: I) {
-        self.push_hash(item_to_hash(item))
+        let build_hasher = H::default();
+        self.push_hash(build_hasher.hash_one(item));
```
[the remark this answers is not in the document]
itsibitzi: 👍 

I've removed this as a default function.

aneubeck (commented):
LGTM

crates/geo_filters/tests/public_api.rs
```diff
@@ -1,3 +1,24 @@
...
+    c1.push(1);
+    c1.push(2);
+
+    let mut c2 = GeoDistinctCount13::default();
+    c2.push(2);
+    c2.push(3);
+
+    let estimated_size = c1.size_with_sketch(&c2);
+    assert!((3.0_f32 * 0.9..=3.0_f32 * 1.1).contains(&estimated_size));
+}
```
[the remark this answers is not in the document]
itsibitzi: Added it using this trick:

```rust
#[doc = include_str!("../README.md")]
#[cfg(doctest)]
pub struct ReadmeDocTests;
```

This seems to need to be in `lib.rs` because there are some issues in the rust tool chain that prevent doc tests from running.

aneubeck (approved):

PR #71, opened 2025-07-23
Basic serialization and deserialization of diff filters

itsibitzi:
# What does this change?

It is useful to be able to serialize the geofilters into a compact byte representation so that they can be stored and compared 
against other filters in the future.

This PR adds a few functions and tests to support writing the geofilters to, and reading geofilters from byte buffers.

Implemented only for `Diff` filters for now.

# How does this work?

The implementation is pretty straight forward if you're familiar with the way the most and least significant bits are split (I wasn't at the beginning of today). Notes on that below...

We serialize the MSB array as numbers and the LSB bitmap as a bitmap (with one byte to represent the unoccupied bits in the most significant `u64`).

I used some unsafe Rust to convert between the various representations. This means this library is not safe if you're sharing the serialized representations across machines with different endiannesses. I've left this in a comment for now. I partially implemented supporting big-endian platforms but given that I won't be able to easily test this I stashed it for now. Maybe someone will one day want a geofilter on an IBM mainframe (a tier 2 rust target!).

I've added a test that throws random values into the filter, writes it into a vec, and then reads it back out again to check that nothing has changed.

## Bitmap representation

The bitfield that represents the spread over the geometric distribution is encoded in two parts. The most significant bits and the last significant bits are held separately in two different representations.

But why?

As items are inserted into the filter they are much more likely to land in the buckets represented by the least significant bits due to the geometric distribution. As you walk up the bit array the sparsity increases, because it's less likely that an item will be hashed to a bucket there. As a result, it doesn't make sense to store a dense representation for the entire filter, as you'd end up with a lot of zeros in the more significant end of the bitmap. 

As a result this crate represents the sparser part of the bitmap efficiently using a vector of numbers representing the occupied indices and represents the denser part of the bitmap with a regular bit array.

# Review notes

I also added some comments for some other bits as I was figuring this data structure out. I left them in this PR because they seem generally useful for someone who is not super familiar with the implementation of the MSB/LSB split. It would be nice to get some confirmation that these comments are correct. And let me know if you think they are not needed. I have been taking notes as I go along understanding this data structure and I plan on writing a doc for non-academic CS people to be able to get a grasp on it's workings, which might make these kinds of comments a bit irrelevant.

Please also let me know if there's any mistakes in my summary of the data structure in this PR description.


crates/geo_filters/src/diff_count/bitvec.rs
```diff
@@ -142,6 +146,60 @@ impl BitVec<'_> {
...
         size_of_val(num_bits) + blocks.len() * size_of::<u64>()
     }
+
+    pub fn from_bytes(mut buf: &[u8]) -> Self {
+        if buf.is_empty() {
+            return Self::default();
+        }
+
+        // The first byte of the serialized BitVec is used to indicate how many
+        // of the bits in the left-most byte are *unoccupied*.
```
aneubeck, line 156: I think this is not quite correct.
It's the right-most "word" (where word = u64).

crates/geo_filters/src/diff_count/serde.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use std::{borrow::Cow, ops::Deref as _};
+
+use crate::{config::GeoConfig, Diff};
+
+use super::{bitvec::BitVec, GeoDiffCount};
+
+impl<'a, C: GeoConfig<Diff>> GeoDiffCount<'a, C> {
+    /// Create a new [`GeoDiffCount`] from a slice of bytes
+    #[cfg(target_endian = "little")]
```
aneubeck, line 21: don't you want to add this macro also to the bitvec functions?

crates/geo_filters/src/diff_count/serde.rs
```diff
@@ -0,0 +1,77 @@
+//! Convert a [`GeoDiffCount`] to and from byte arrays.
```
aneubeck, line 1: If you want to keep the serialization in a separate file, then let's not call it serde...
We don't implement anything serde specific in here.

crates/geo_filters/src/diff_count/serde.rs
```diff
@@ -0,0 +1,77 @@
...
+            msb: Cow::Borrowed(msb),
+            lsb: BitVec::from_bytes(&buf[msb_bytes_len..]),
+        }
+    }
+
+    /// Create a new [`GeoDiffCount`] from a slice of bytes
+    #[cfg(target_endian = "big")]
+    pub fn from_bytes(c: C, buf: &'a [u8]) -> Self {
+        unimplemented!("not supported on big endian platforms")
+    }
```
aneubeck, line 52: why even define the function on these platforms?

crates/geo_filters/src/diff_count/bitvec.rs
itsibitzi: Ah good catch. Typo :(

crates/geo_filters/src/diff_count/serde.rs
itsibitzi: I was just meaning as a contraction for serialization/deserialization. But will rename!

crates/geo_filters/src/diff_count.rs
```diff
@@ -229,15 +237,19 @@ impl<C: GeoConfig<Diff>> GeoDiffCount<'_, C> {
                 Err(idx) => {
                     msb.insert(idx, bucket);
                     if msb.len() > self.config.max_msb_len() {
+                        // We have too many values in the MSB sparse index vector,
+                        // let's move the smalles MSB value into the LSB bit vector
```
aneubeck, line 241:
```suggestion
                        // let's move the smallest MSB value into the LSB bit vector
```

crates/geo_filters/src/diff_count.rs
```diff
@@ -580,4 +594,67 @@ mod tests {
...
+            // Select a random number of items to insert
+            let items = (1..1000).choose(&mut rnd).unwrap();
+
+            for _ in 0..items {
+                before.push_hash(rnd.next_u64());
+            }
+
+            let mut writer = vec![];
+
+            // Insert some padding to emulate alignment issues with the slices
```
aneubeck, line 633:
```suggestion
            // Insert some padding to emulate alignment issues with the slices.
```

crates/geo_filters/src/diff_count/serde.rs
itsibitzi: I've actually just moved it into the main file since I'm abandoning big endian for now.

crates/geo_filters/src/diff_count/bitvec.rs
```diff
@@ -142,6 +146,62 @@ impl BitVec<'_> {
         let Self { num_bits, blocks } = self;
         size_of_val(num_bits) + blocks.len() * size_of::<u64>()
     }
+
+    #[cfg(target_endian = "little")]
+    pub fn from_bytes(mut buf: &[u8]) -> Self {
+        if buf.is_empty() {
+            return Self::default();
+        }
+
```
aneubeck, line 155: can you remove empty lines here as well?

crates/geo_filters/src/diff_count.rs
```diff
@@ -208,16 +209,23 @@ impl<C: GeoConfig<Diff>> GeoDiffCount<'_, C> {
...
             self.lsb.toggle(bucket.into_usize());
         } else {
             let msb = self.msb.to_mut();
             match msb.binary_search_by(|k| bucket.cmp(k)) {
                 Ok(idx) => {
+                    // The bit is already set in the MSB sparse bitset, remove it (XOR)
                     msb.remove(idx);
+
+                    // We have removed a value from our MSB, move a value in the
+                    // LSB into the MSB
```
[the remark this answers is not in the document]
itsibitzi: Opened a PR to address this: https://github.com/github/rust-gems/pull/72

I didn't add a new test but I added a way of deterministically replaying fuzz tests. I'm not sure how we'd reliably get this to fail other than keep fuzzing until we find a failure and then hardcode the seed? Maybe by using a no-op hash function (identity function)?

PR #72, opened 2025-07-24
Deterministic RNG test harness and address bug in `BitVec`

itsibitzi:
## Why?

There have been random test failures caused by some invariants not being maintained in the LSB `BitVec`. This presents two problems.

1. There seem to be bugs in our code, at least according to the tests.
2. The tests are random and thus failures are hard to reproduce

## How?

To address this I've done two things.

Firstly, I implement a test harness for running tests which require a random number generator (RNG). This harness accepts a lambda function which has a single argument, a RNG. Under normal conditions the RNG is seeded randomly as a typical thread local RNG. The difference occurs when you get a test failure, at which point the harness captures the panic and prints out the RNG seed. You can then rerun your tests with `TEST_SEED=...` set as an environment variable in order to start the test with the seed that failed. This allows developers to rerun the test and see the failure immediately, a generally useful thing to be able to do :)

E.g. you run some tests with `cargo test` and get the following error message from a flakey test:

```
---- diff_count::tests::test_xor_plus_mask stdout ----

thread 'diff_count::tests::test_xor_plus_mask' panicked at crates/geo_filters/src/diff_count.rs:535:21:
assertion `left == right` failed
  left: ~12.37563 (msb: [419, 328, 284, 251, 237, 180, 148, 117, 108, 67, 60, 26], |lsb|: 26)
 right: ~12.37563 (msb: [419, 328, 284, 251, 237, 180, 148, 117, 108, 67, 60, 26], |lsb|: 0)
Test failed! Reproduce with: TEST_SEED=16501082297235867130
```

You can then re-run with `TEST_SEED=16501082297235867130 cargo test` and the test will fail reliably.

I rewrote all our `tests` to use this since it's good to be able to reproduce issues. I did not roll it out to the `evaluation` code.

Secondly, I fix issue where `BitVec` was required to be structurally identical rather than just semantically equivalent to pass equality checks. Specifically, an empty `BitVec` with no blocks is semantically the same as a `BitVec` with `>=1` block where all the bits are zeros, but the `PartialEq` function didn't consider them the same.

This caused issues when bits in the LSB were toggled directly. An item is pushed, hashed and assigned a bucket. If the bucket is in the MSB range then the LSB is updated and potentially resized. If the bucket is in the LSB range the bit is toggled but the bucket is _not_ resized. This is because we expect most items to toggle bits in the LSB range so not resizing the buffer every time this happen saves us on some performance.

This issue would only happen relatively rarely, given that it needed the last item inserted to toggle the last bit in the LSB before the equality check is performed. Usually there is more than 1 bit in the LSB set before the final item is inserted.

I addressed this issue by changing how `PartialEq::eq` is implemented to check that the set bits are equivalent using the existing `bit_chunks` iterator.

I think it in the serialization code (PR not yet merged) it would make sense to resize the LSB before serialization to prevent writing empty bytes to disk.

## Alternative approaches

For the deterministic RNG I could have used something like `proptest`, but it felt like too big/powerful a dependency for this limited use case.

For the `BitVec` stuff....

Rather than make the `Eq` function check the semantic equivalence we could spend more effort maintaining the structural equivalence of the `BitVec`.

The first thing I tried was to make toggling the LSB check the size of the vectors and resize if needed - this fixed the issue, but was bad for performance.

I tried changing how the LSB is updated when the MSB has an insert, but this caused quite a few knock on effects due to the fact that the most-significant-bit in the LSB range is truncated when it's resized. Removing this resize code had a load of knock-on issues that, to me, seemed harder to solve.

If anyone feels strongly that the `BitVec` should do it's comparisons based on the specific state and that the semantic equivalence is not good enough then we can rework this PR.

### DIFF
--- a/crates/geo_filters/src/config/lookup.rs
+++ b/crates/geo_filters/src/config/lookup.rs
@@ -45,9 +45,12 @@ impl HashToBucketLookup {
 
 #[cfg(test)]
 mod tests {
-    use rand::{RngCore, SeedableRng};
+    use rand::RngCore;
 
-    use crate::config::{hash_to_bucket, phi_f64};
+    use crate::{
+        config::{hash_to_bucket, phi_f64},
+        test_rng::prng_test_harness,
+    };
 
     use super::HashToBucketLookup;
 
@@ -66,15 +69,17 @@ mod tests {
     fn lookup_random_hashes_variance<const B: usize>(n: u64) -> f64 {
         let phi = phi_f64(B);
         let buckets = HashToBucketLookup::new(B);
-        let mut var = 0.0;
-        let mut rnd = rand::rngs::StdRng::from_os_rng();
-        for _ in 0..n {
-            let hash = rnd.next_u64();
-            let estimate = buckets.lookup(hash) as f64;
-            let real = hash_to_bucket(phi, hash) as f64;
-            let err = estimate - real; // assume the mean = 0.0
-            var += err.powf(2.0) / n as f64;
-        }
-        var
+
+        prng_test_harness(|rnd| {
+            let mut var = 0.0;
+            for _ in 0..n {
+                let hash = rnd.next_u64();
+                let estimate = buckets.lookup(hash) as f64;
+                let real = hash_to_bucket(phi, hash) as f64;
+                let err = estimate - real; // assume the mean = 0.0
+                var += err.powf(2.0) / n as f64;
+            }
+            var
+        })
     }
 }
--- a/crates/geo_filters/src/diff_count.rs
+++ b/crates/geo_filters/src/diff_count.rs
@@ -95,16 +95,14 @@ impl<C: GeoConfig<Diff>> GeoDiffCount<'_, C> {
     /// having to construct another iterator with the remaining `BitChunk`s.
     fn from_bit_chunks<I: Iterator<Item = BitChunk>>(config: C, chunks: I) -> Self {
         let mut ones = iter_ones::<C::BucketType, _>(chunks.peekable());
-
         let mut msb = Vec::default();
         take_ref(&mut ones, config.max_msb_len() - 1).for_each(|bucket| {
             msb.push(bucket);
         });
         let smallest_msb = ones
             .next()
-            .map(|bucket| {
-                msb.push(bucket);
-                bucket
+            .inspect(|bucket| {
+                msb.push(*bucket);
             })
             .unwrap_or_default();
 
@@ -365,6 +363,7 @@ mod tests {
     use crate::{
         build_hasher::UnstableDefaultBuildHasher,
         config::{iter_ones, tests::test_estimate, FixedConfig},
+        test_rng::prng_test_harness,
     };
 
     use super::*;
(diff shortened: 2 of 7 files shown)


crates/geo_filters/src/diff_count.rs
```diff
@@ -95,16 +95,14 @@ impl<C: GeoConfig<Diff>> GeoDiffCount<'_, C> {
...
         take_ref(&mut ones, config.max_msb_len() - 1).for_each(|bucket| {
             msb.push(bucket);
         });
         let smallest_msb = ones
             .next()
-            .map(|bucket| {
-                msb.push(bucket);
-                bucket
+            .inspect(|bucket| {
+                msb.push(*bucket);
```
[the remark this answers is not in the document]
itsibitzi: The original code didn't modify the value?

crates/geo_filters/src/config/lookup.rs
```diff
@@ -66,15 +69,17 @@ mod tests {
...
-        for _ in 0..n {
-            let hash = rnd.next_u64();
-            let estimate = buckets.lookup(hash) as f64;
-            let real = hash_to_bucket(phi, hash) as f64;
-            let err = estimate - real; // assume the mean = 0.0
-            var += err.powf(2.0) / n as f64;
-        }
-        var
+
+        prng_test_harness(|rnd| {
```
aneubeck, line 73: why is it useful to wrap this for loop into the harness?
The loop cannot really panic, so we won't ever print the seed of the random number generator, right?

crates/geo_filters/src/test_rng.rs
```diff
@@ -0,0 +1,25 @@
+use std::panic::{catch_unwind, resume_unwind, AssertUnwindSafe};
+
+use rand::{rngs::StdRng, SeedableRng as _};
+
+/// Provides a seeded random number generator to tests which require some
+/// degree of randomization. If the test panics the harness will print the
+/// seed used for that run. You can then pass in this seed using the `TEST_SEED`
+/// environment variable when running your tests.
+pub fn prng_test_harness<F, T>(test_fn: F) -> T
```
aneubeck, line 9: I think it would make sense to also get a parameter `n` for the number of iterations the function should be executed.
At least that's the usual scenario in which this function is used.
This has also the advantage that you won't have to run the function 1000 times if TEST_SEED env is set :)

crates/geo_filters/src/test_rng.rs
```diff
@@ -0,0 +1,25 @@
+use std::panic::{catch_unwind, resume_unwind, AssertUnwindSafe};
```
aneubeck, line 1: nice!

crates/geo_filters/src/config/lookup.rs
itsibitzi: I'll make it so this function accepts an RNG rather than making a new one and lift the harness to the caller functions.

crates/geo_filters/src/diff_count/bitvec.rs
```diff
@@ -37,15 +36,7 @@ impl BitVec<'_> {
...
-    ///
-    /// NOTE: If the bitchunks iterator is empty, the result is NOT sized to `num_bits` but will
-    ///       be EMPTY instead.
-    pub fn from_bit_chunks<I: Iterator<Item = BitChunk>>(
-        mut chunks: Peekable<I>,
-        num_bits: usize,
-    ) -> Self {
-        // if there are no chunks, we keep the size zero
-        let num_bits = chunks.peek().map(|_| num_bits).unwrap_or_default();
+    pub fn from_bit_chunks<I: Iterator<Item = BitChunk>>(chunks: I, num_bits: usize) -> Self {
```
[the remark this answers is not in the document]
itsibitzi: There weren't any for the `BitVec` but I did find some for other functions 😅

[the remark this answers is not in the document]
itsibitzi: This is because `BitChunksOnes` iterator already has peekable iterators internally. They're required for `next`, but not required externally, so the `Peekable` wrapper here in the `from_bit_chunks` function is not needed.

PR #73, opened 2025-07-25
Ensure that BitVec has a deterministic size

aneubeck:
There was some "unexpected" behaviour in BitVec.
It took as argument a size, but then replaced it (sometimes) with the larger bit position?!?
I removed that hack and fixed the calling places.


itsibitzi:
Following a discussion on Slack with Alex I've `cherry-pick`ed these changes onto my branch. Going to close this PR.

PR #80, opened 2025-08-11
Implement a novel consistent hashing algorithm with replication

aneubeck:
This algorithm is special, since it's complexity is not dependent on the number of nodes!
The simple implementation is quadratic in the number of desired replication nodes. An r * log(r) complexity could be achieved by utilizing a heap or sorted set data structure. Since r is typically small, this overhead is not worth it though.

[rendered](https://github.com/github/rust-gems/blob/aneubeck/sampling/crates/consistent-choose-k/README.md)


crates/consistent-hashing/README.md
```diff
@@ -0,0 +1,60 @@
+# Consistent Hashing
+
+Consistent hashing maps keys to a changing set of nodes (shards, servers) so that when nodes join or leave, only a small fraction of keys move. It is used in distributed caches, databases, object stores, and load balancers to achieve scalability and high availability with minimal data reshuffling.
+
+Common algorithms
+- [Consistent hashing](https://en.wikipedia.org/wiki/Consistent_hashing) (hash ring with virtual nodes)
+- [Rendezvous hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing)
+- [Jump consistent hash](https://en.wikipedia.org/wiki/Jump_consistent_hash)
```
itsibitzi, line 8: Wikipedia link didn't work for me, maybe this: https://arxiv.org/pdf/1406.2294 ?

crates/consistent-hashing/README.md
```diff
@@ -0,0 +1,99 @@
...
+`consistent_choose_k` makes `k` calls to `consistent_choose_max` which calls `consistent_hash` another `k` times.
+In total, `consistent_hash` is called `k * (k+1) / 2` Utilizing a `O(1)` solution for `consistent_hash` leads to a `O(k^2)` runtime.
+This runtime can be further improved by replacing the max operation with a heap where popped elements are updated according to the new arguments `n` and `k`.
+With this optimization, the complexity reduces to `O(k log k)`.
+With some probabilistic bucketing strategy, it should be possible to reduce the expected runtime to `O(k)`.
+For small `k` neither optimization is probably improving the actual performance though.
+
+The next section proves the correctness of this algorithm.
+
+## N-Choose-R replication
```
itsibitzi, line 70: Should this title be "N-Choose-K"?

crates/consistent-hashing/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+            for k in 0..=i {
+                while samplers[k].peek() >= Some(&(n - k)) && n - k > 0 {
+                    samplers[k].next();
+                }
+                max = max.max(samplers[k].peek().unwrap() + k);
+            }
+            samples.push(max);
+            n = max;
+        }
+        samples.sort();
```
itsibitzi, line 251: Is it important that we sort this list within our function? If the user wants the list sorted they could do it themselves?

itsibitzi (commented):
An initial batch of comments.

crates/consistent-hashing/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+impl<H: ManySeqBuilder> ConsistentChooseKHasher<H> {
+    pub fn new(builder: H, k: usize) -> Self {
+        Self { builder, k }
+    }
+
+    // TODO: Implement this as an iterator!
+    pub fn prev(&self, mut n: usize) -> Vec<usize> {
+        let mut samples = Vec::with_capacity(self.k);
+        let mut samplers: Vec<_> = (0..self.k)
+            .map(|i| ConsistentHashRevIterator::new(n - i, self.builder.seq_builder(i)).peekable())
```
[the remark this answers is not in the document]
aneubeck: added another function which takes a vector as input.
Changed the implementation to get rid of the second vector.
Downside is that we have to do more work now in order to fetch the next consistent hash for a specific sampler 🤷 

Generally, performance is actually not so bad...

jorendorff (approved):
OK, if I was going to make one change to this code, I'd add one comment at the top of `consistent_hash` or on `ConsistentHashRevIterator::next` saying "this is Algorithm 5 from [JumpBackHash]". Would have saved me some time!

But no point causing conflicts, if we're just going to delete it.

PR #93, opened 2026-01-14
Add famst crate: Fast Approximate Minimum Spanning Tree

jorendorff:
Implementation of the FAMST algorithm from Almansoori & Telek (2025).

Features:
- Generic over data type T and distance function
- Uses NN-Descent for ANN graph construction
- Three-phase approach: ANN graph, component connection, edge refinement
- O(dn log n) time complexity, O(dn + kn) space complexity

Paper: https://arxiv.org/abs/2507.14261

(Entirely vibe-coded by Opus 4.5.)

### DIFF
--- a/crates/famst/src/lib.rs
+++ b/crates/famst/src/lib.rs
@@ -10,6 +10,29 @@
 //! 3. Iterative edge refinement
 //!
 //! Generic over data type `T` and distance function.
+//! 
+//! While https://github.com/cmuparlay/ParlayANN is not specifically for MST construction,
+//! it gives some idea how far one can scale ANN graph construction.
+//! 
+//! One problem with very large graphs are cache misses. To address them, a common procedure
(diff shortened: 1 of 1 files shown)


crates/famst/src/lib.rs
```diff
@@ -0,0 +1,1026 @@
...
+{
+    let n = data.len();
+    let k = config.k.min(n - 1);
+
+    if k == 0 || n <= 1 {
+        return AnnGraph::new(vec![vec![]; n], vec![vec![]; n]);
+    }
+
+    // Initialize with random neighbors using max-heap for each point
+    let mut heaps: Vec<BinaryHeap<NeighborEntry>> = Vec::with_capacity(n);
```
aneubeck: Can't we allocate one gigantic vector and slice it instead of creating millions of tiny heaps?

PR #94, opened 2026-01-16
Open source HRIBLT 

itsibitzi:


crates/hriblt/docs/sizing.md
```diff
@@ -0,0 +1,17 @@
...
+
+Regardless, there are some advantages to getting the size of your decoding session correct the first time. An example might be if you're performing set reconciliation over some RPC and you want to minimise the number of round trips it takes to perform a decode.
+
+## Coded Symbol Multiplier
+
+The number of coded symbols required to find the difference between two sets is proportional to the difference between the two sets. The following chart shows the relationship between the number of coded symbols required to decode HRIBLT and the size of the diff. Note that the size of the base set (before diffs were added) was fixed.
+
+`y = len(coded_symbols) / diff_size`
+
+![Coded symbol multiplier](./assets/coded-symbol-multiplier.png)
```
aneubeck: We also need to visualize the standard deviation (or maybe percentiles like 90%-ile, 99%-ile). Max doesn't really make sense, since in theory it is infinite.

Those are almost more important than the average.

crates/hriblt/README.md
```diff
@@ -0,0 +1,55 @@
...
+## Usage
+
+Add the library to your `Cargo.toml` file.
+
+```toml
+[dependencies]
+hriblt = "0.1"
+```
+
+Create two encoding sessions, one containing your data, and another containing the counter-parties data. This counterparty data might have been sent to you over a network for example.
```
aneubeck: replace your and counter-party with Alice and Bob from the start...

crates/hriblt/src/coded_symbol.rs
```diff
@@ -0,0 +1,127 @@
...
+fn indices_contains<T: std::hash::Hash>(
+    state: &impl HashFunctions<T>,
+    value: &T,
+    stream_len: usize,
+    i: usize,
+) -> i32 {
+    if stream_len > 32 && i % 32 != 0 {
+        let seed = i % 4;
+        let j = index_for_seed(state, value, stream_len, seed as u32);
+        if i == j { 1 } else { 0 }
```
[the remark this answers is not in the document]
itsibitzi: https://godbolt.org/z/6bdEdKP54

I made the above to see if LLVM would keep the branch. It looks like `foo` and `bar` both get compiled to the same instructions in `-O3` to the point where the compiler only keeps `bar`?

I've moved it to use `i32::from(i == j)` since this is less ambiguous from the compiler.

crates/hriblt/evaluation/bench.rs
```diff
@@ -0,0 +1,303 @@
...
+        }
+    }
+
+    let set_b: Vec<u64> = set_b.into_iter().collect();
+
+    // Create encoding sessions for both sets with max capacity
+    let state = DefaultHashFunctions;
+
+    let mut encoder_a = EncodingSession::new(state, 0..max_symbols);
+    encoder_a.extend(set_a.iter().copied());
```
aneubeck: this whole setup is crazily inefficient especially if you want to compute it for ALL diff sizes within a range.
We should at least change this code to use exponentially growing diff sizes.
On top, one doesn't need TWO sets... it is totally fine to use a single set with n elements.
If you want to be fancy, then you can insert some of the n elements as insertions and the others as deletions.

With this procedure, one can simply compute the next encodingsession by adding another (distinct) element to it. The distinct aspect can be achieved by starting off with a random set of unique integers (in random order). At every x-th integer addition/removal you can then try to decode the set.

crates/hriblt/evaluation/bench.rs
```diff
@@ -0,0 +1,303 @@
...
+                eprintln!("Trial {} failed (diff_size={})", i + 1, diff_size);
+            }
+        }
+    }
+
+    eprintln!();
+    eprintln!("Results:");
+    eprintln!("  Successes: {}/{} ({:.1}%)", successes, args.trials, 
+             100.0 * successes as f64 / args.trials as f64);
+    eprintln!("  Failures:  {}/{} ({:.1}%)", failures, args.trials,
```
aneubeck: have there been failures?

crates/hriblt/evaluation/README.md
```diff
@@ -0,0 +1,78 @@
...
+- **Gray region**: Full range (p0 to p99) - min to ~max overhead observed
+- **Blue region**: Interquartile range (p25 to p75) - where 50% of trials fall
+- **Dark line**: Median (p50) - typical overhead
+
+A lower overhead means more efficient encoding. An overhead of 1.0x would mean the number of coded symbols exactly equals the diff size (theoretical minimum).
+
+Typical results show:
+- Small diff sizes (1-10) have higher variance and overhead due to the probabilistic nature of the algorithm
+- Larger diff sizes (50+) converge to a more stable overhead around 1.3-1.5x
+- The algorithm successfully decodes 100% of trials when given up to 10x the diff size in coded symbols
```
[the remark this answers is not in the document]
aneubeck: I would actually expect that we need AT MOST 50% overhead after some diff size.

PR #107, opened 2026-04-29
First part of a HashSortedMap

aneubeck:
In a subsequent step, merging functions and sorted iterators will be added.


crates/hash-sorted-map/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod group_ops;
+pub mod hash_sorted_map;
```
jorendorff:
```suggestion
mod group_ops;
mod hash_sorted_map;

pub use hash_sorted_map::{Entry, HashSortedMap, OccupiedEntry, VacandEntry};
```

crates/hash-sorted-map/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+`HashSortedMap<K, V, S>` is a Swiss-table-inspired hash map that uses:
+
+- **Overflow chaining** instead of open addressing — groups that fill up link
+  to overflow groups rather than probing into neighbours.
+- **Slot hint** — a preferred slot index derived from the hash, checked before
+  scanning the group. Gives a direct hit on most inserts at low load.
+- **SIMD group scanning** — uses NEON on aarch64, SSE2 on x86\_64, and a
+  scalar fallback elsewhere to scan 8–16 control bytes in parallel.
+- **AoS group layout** — each group stores its control bytes, keys, and values
```
jorendorff, line 36: Suggest linking Wikipedia here: https://en.wikipedia.org/wiki/AoS_and_SoA#Array_of_structures_of_arrays

Wikipedia calls this AoSoA (since each Group is a struct of arrays) or "tiled AoS"

crates/hash-sorted-map/src/hash_sorted_map.rs
```diff
@@ -0,0 +1,753 @@
...
+    }
+}
+
+impl<K, V, S> HashSortedMap<K, V, S> {
+    pub fn with_hasher(hash_builder: S) -> Self {
+        Self::with_capacity_and_hasher(0, hash_builder)
+    }
+
+    pub fn with_capacity_and_hasher(capacity: usize, hash_builder: S) -> Self {
+        let adjusted = capacity.checked_mul(8).unwrap_or(usize::MAX) / 7;
```
jorendorff: I think we need to adjust more than that, because the expected load factor is not as high as hashbrown.

I did a few [simulations](https://gist.github.com/jorendorff/ed8b95374e1422b365b5b3b9f7d40e03) and we need to multiply by about 1.30 here, or 1.42 if GROUP_SIZE is 8. That is assuming `alloc_groups` overallocates by a further 1/8 and that we want `with_capacity` to be an honest effort to accommodate that many entries without growing.

jorendorff (commented):
This is cool.

Can I see the merge/iteration operations you have in mind, before stamping this?

crates/hash-sorted-map/src/hash_sorted_map.rs
aneubeck: 👍 
(I had copilot compute those numbers as well, or at least something similar, just forgot to change the code)

aneubeck:
Here is the very simple sorting of each group:
https://github.com/github/blackbird/blob/aneubeck/reindex4/crates/index/src/indexer/prefix_map.rs#L288
and here the sorted iterator:
https://github.com/github/blackbird/blob/aneubeck/reindex4/crates/index/src/indexer/prefix_map.rs#L390

In the ported version, we will need to introduce a new type, since the hash map will be broken after the sorting.
But one might want to iterate multiple times.

Merging is implemented here:
https://github.com/github/blackbird/blob/aneubeck/reindex4/crates/index/src/indexer/memory_index.rs#L191-L204

We could provide a helper function in the rust crate for this.

Feel free to merge/make changes, since I will be OOO tomorrow.

crates/hash-sorted-map/src/hash_sorted_map.rs
```diff
@@ -0,0 +1,753 @@
...
+            if empty_mask != 0 {
+                hint = group_ops::lowest(empty_mask);
+                break;
+            }
+            let overflow = group.overflow;
+            if overflow != NO_OVERFLOW {
+                group = &mut self.groups[overflow as usize];
+            } else {
+                let new_gi = self.num_groups as usize;
+                self.groups[gi].overflow = new_gi as u32;
```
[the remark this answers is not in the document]
aneubeck: doesn't work, since we need to fetch the len of the group which doesn't work when one item is already borrowed :(

crates/hash-sorted-map/src/hash_sorted_map.rs
```diff
@@ -0,0 +1,775 @@
...
+
+/// View into an occupied entry.
+pub struct OccupiedEntry<'a, V> {
+    value: &'a mut V,
+}
+
+/// View into a vacant entry. Holds the borrow of the map plus the hash, key,
+/// and pre-computed insertion slot.
+pub struct VacantEntry<'a, K, V, S> {
+    map: &'a mut HashSortedMap<K, V, S>,
```
jorendorff: OccupiedEntry is fine, but modifying memory directly owned by this HashSortedMap from another mut pointer (group/tail), while this mut reference exists, is UB.

The annoying workaround is like

```suggestion
    phantom: PhantomData<&'a mut HashSortedMap<K, V, S>>,
    map: *mut HashSortedMap<K, V, S>,
```

and only make mut references via one of the two pointers at a time, which means you have to alternate.

crates/hash-sorted-map/src/hash_sorted_map.rs
```diff
@@ -0,0 +1,775 @@
...
+        unsafe {
+            group.keys[hint]
+                .as_mut_ptr()
+                .copy_from_nonoverlapping(key_src, 1)
+        };
+        unsafe {
+            group.values[hint]
+                .as_mut_ptr()
+                .copy_from_nonoverlapping(value_src, 1)
+        };
```
jorendorff: Style nit.

```suggestion
        unsafe {
            group.keys[hint]
                .as_mut_ptr()
                .copy_from_nonoverlapping(key_src, 1);
            group.values[hint]
                .as_mut_ptr()
                .copy_from_nonoverlapping(value_src, 1);
        }
```

crates/hash-sorted-map/src/hash_sorted_map.rs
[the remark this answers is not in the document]
jorendorff: The bug is real though - I pushed a test.

crates/hash-sorted-map/OPTIMIZATIONS.md
```diff
@@ -0,0 +1,176 @@
...
+
+HashSortedMap checks a preferred slot before scanning the group:
+```rust
+let hint = slot_hint(hash);  // 3 bits from hash → slot index
+if ctrl[hint] == EMPTY { /* direct insert */ }
+if ctrl[hint] == tag && keys[hint] == key { /* direct hit */ }
+```
+
+hashbrown does **not** have this optimization — it always does a full SIMD
+group scan. At ~50% load, the hint hits ~58% of the time, avoiding the scan
```
jorendorff: How is this possible? If the key being inserted has a random hash, then at X% load it should hit exactly X% of the time.

crates/hash-sorted-map/OPTIMIZATIONS.md
```diff
@@ -0,0 +1,176 @@
+# HashSortedMap vs. Rust Swiss Table (hashbrown): Optimization Analysis
+
+## Executive Summary
```
jorendorff, line 3: AI slop. :P It's honestly better than nothing, but I'd be happier if you did a read through this and just deleted everything that's not something you would say as a human.

crates/hash-sorted-map/OPTIMIZATIONS.md
```diff
@@ -0,0 +1,176 @@
...
+| m/8 (12.5%, default) | 8.04 |
+| m/4 (25%) | 8.33 |
+| m/2 (50%) | 8.93 |
+| m/1 (100%) | 10.31 |
+| 0 (grow immediately) | 6.96 |
+
+**Conclusion**: Smaller reserves are faster — growing early is cheaper than
+traversing overflow chains. The `m/8` default implicitly enforces ~62.5% max
+load, which aligns with the mathematical analysis (Poisson model, 3σ
+confidence).
```
jorendorff: ...What mathematical analysis? Is that load number correct? It's different from what the code says.

jorendorff (approved):

crates/hash-sorted-map/src/hash_sorted_map.rs
aneubeck: I have no idea whether I did the right thing now... I simply added (*map) dereferencing code to the insert implementation

aneubeck: should insertion also have the same lifetime (via a phantomdata)?

crates/hash-sorted-map/OPTIMIZATIONS.md
aneubeck: remove the awkward explanation from copilot and added a more likely reason for the difference.

crates/hash-sorted-map/OPTIMIZATIONS.md
aneubeck: Remove the mathematical explanation :)

PR #112, opened 2026-05-08
Implement sorting by hash

aneubeck:
This PR
* adds iterators
* adds sort_by_hash feature that data is iterated according to their hash
* adds benchmarks for sort/merge use cases
Note: hashmap remains intact after sorting now.


crates/hash-sorted-map/OPTIMIZATIONS.md
```diff
@@ -106,17 +105,32 @@ the overflow path.
...
+Originally, HashSortedMap checked a preferred slot before scanning the group:
 ```rust
 let hint = slot_hint(hash);  // 3 bits from hash → slot index
 if ctrl[hint] == EMPTY { /* direct insert */ }
 if ctrl[hint] == tag && keys[hint] == key { /* direct hit */ }
 ```
 
-hashbrown does **not** have this optimization — it always does a full SIMD
-group scan. The reason why the performance is different is probably due to the different overflow strategies and the different load factors.
+**Experimental finding**: This scalar check **hurts performance** on random
```
jorendorff, line 117: This makes so much sense. And removing it from lookup paths explains why we can sort the map and it's still a map. Pretty great outcome!

How valuable is it for resizing? Even if it usually hits, surely it's equally fast to find the first empty slot in the group with SIMD, and that will hit even more often.

crates/hash-sorted-map/README.md
```diff
@@ -42,45 +42,24 @@ keys, which means:
...
-- **Growth is ~2× faster** than hashbrown thanks to the optimized
-  `insert_for_grow` path that skips duplicate checking and uses raw copies.
-- The remaining gap to FoldHashMap (~11%) comes from foldhash's extremely
-  efficient hash function that pipelines well with hashbrown's SIMD scan.
+Latest local Criterion snapshot from this repository's
+`target/criterion` outputs (lower is better):
+
+| Scenario                                     | HashSortedMap | Comparison                             | Result      |
+| :------------------------------------------- | ------------: | :------------------------------------- | :---------- |
+| Insert 1000 trigrams (pre-sized)             |       7.34 µs | hashbrown::HashMap: 12.88 µs           | ~43% faster |
```
jorendorff: This doesn't agree with the other file, which puts us 32% slower than hashbrown on the same microbenchmark. Different architecture?

crates/hash-sorted-map/src/hash_sorted_map.rs
```diff
@@ -541,42 +624,18 @@ impl<'a, K: Hash + Eq, V, S: BuildHasher> VacantEntry<'a, K, V, S> {
...
-/// (`insert_for_grow` relies on the same assumption to skip its own
-/// capacity check.)
+/// grow before inserting. Grows the map, then re-walks via `entry()` to find
+/// the new insertion slot.
 #[cold]
 #[inline(never)]
 fn insert_after_grow<K: Hash + Eq, V, S: BuildHasher>(
     map: &mut HashSortedMap<K, V, S>,
-    hash: u64,
+    _hash: u64,
```
jorendorff: Clean this up?

crates/hash-sorted-map/README.md
jorendorff: It would be best to have all the benchmarks in one place and up to date, and preferably all on Intel if that's what we think most cloud servers have.

jorendorff (approved):

crates/hash-sorted-map/OPTIMIZATIONS.md
aneubeck: changed the grow code as well. As a result all occupied slots are at the beginning, so no more special treatment when sorting...

crates/hash-sorted-map/README.md
aneubeck: reran on codespace intel CPU.
runs vary sometimes pretty drastically and generally the xeon CPUs are much slower than my M4 machine :(

We might want to retest on a dedicated production machine at some point

PR #113, opened 2026-05-08
Add sparse gram extraction

aneubeck:
Here is a highly optimized sparse-ngrams extraction implementation.
It limits the maximum size of grams to 8 in this version without the option to configure.

In subsequent PRs, I will add functions for extracting grams for a query string.
It's even more interesting to extract sparse ngrams for a regular expression...


crates/sparse-ngrams/src/ngram.rs
```diff
@@ -0,0 +1,132 @@
...
+            let end_hash = prefix_hashes[idx & (MAX_SPARSE_GRAM_SIZE as usize - 1)]
+                .wrapping_mul(POLY_HASH_PRIME)
+                .wrapping_add(content[idx] as u32);
+            // Check the bigram content[idx-1..idx+1]
+            let rolling_hash = end_hash
+                .wrapping_sub(prefix_hashes[(idx - 1) & (MAX_SPARSE_GRAM_SIZE as usize - 1)]
+                    .wrapping_mul(POLY_POWERS[2]));
+            let rolling = NGram::from_rolling_hash(rolling_hash, 2);
+            let direct = NGram::from_bytes(&content[idx - 1..idx + 1]);
+            assert_eq!(rolling, direct, "mismatch at idx={idx}");
```
jorendorff, line 136: It's odd that this test doesn't test ngrams of all sizes.

crates/sparse-ngrams/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+
+mod deque;
+mod extract;
+mod ngram;
+mod table;
+
+pub use ngram::NGram;
+
+/// Number of high-frequency bigrams used to build the priority table.
+pub const NUM_FREQUENT_BIGRAMS: usize = 65534;
```
jorendorff, line 41: Should this be 65536 (2^16)? It doesn't seem to matter either way, as there aren't that many bigrams after case-folding, but I'm curious why it's like this.

crates/sparse-ngrams/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+
+/// Maximum length (in bytes) of a sparse n-gram.
+pub const MAX_SPARSE_GRAM_SIZE: u32 = 8;
```
jorendorff: This seems to be cast to `usize` almost everywhere it's used. Maybe it should just be a `usize`?

crates/sparse-ngrams/src/table.rs
```diff
@@ -0,0 +1,41 @@
...
+//! extraction algorithm to decide where n-gram boundaries fall.
+
+use std::sync::OnceLock;
+
+use crate::NUM_FREQUENT_BIGRAMS;
+
+/// The bigrams in this string are sorted by how frequently they occur in code (descending).
+/// Bigrams are separated by null bytes. Only the first [`NUM_FREQUENT_BIGRAMS`] entries
+/// receive nonzero priority; all other byte pairs default to 0.
+static BIGRAMS_STR: &str = include_str!("bigrams.bin");
```
jorendorff, line 13: Maybe this is a dumb comment, but it seems like this could be 1/3 shorter because the nul bytes aren't necessary.

crates/sparse-ngrams/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+## How it works
+
+Each consecutive byte pair (bigram) is assigned a frequency-based priority from a precomputed table. An n-gram boundary occurs wherever a bigram has lower priority than all bigrams between it and the previous boundary. This is computed efficiently using a monotone deque or a scan-based approach.
+
+For a document of N bytes, this produces at most 3(N−1) n-grams: N−1 bigrams, plus up to 2(N−1) algorithmically selected longer n-grams (up to 8 bytes).
+
+### Selection criterion
+
+A substring of length 3–8 is emitted as a sparse n-gram if and only if every interior bigram priority is strictly greater than the maximum of the left and right boundary bigram priorities.
```
jorendorff, line 23: This section seems redundant.

There was a github blog post about this that I think is worth linking. There's also the comment in `sparse_grams.rs` which is a gentler introduction than this, but I won't insist.

crates/sparse-ngrams/src/ngram.rs
```diff
@@ -0,0 +1,132 @@
+//! Compact n-gram representation using a polynomial rolling hash.
```
jorendorff, line 1: 8-byte n-grams fit in a u64 without hash computation or collisions. I guess we're hashing because we're memory-bound here?

jorendorff (approved):
Looks good.

This is somewhat different from what we do in Blackbird. Without getting into too much detail, we generate more n-grams.

Generally, there's an engineering tradeoff around how many to collect. More is good because queries produce fewer and rarer n-grams; but bad because indexing is more expensive and indexes are larger. It would be nice to let users choose their preferred tradeoff point, i.e. choose their desired ratio of output n-grams per byte of content. I don't know an elegant design for this, though. What we're doing in Blackbird isn't exactly elegant :)

crates/sparse-ngrams/src/lib.rs
aneubeck: left a comment...
mostly because some algorithms require some MAX value and all excluded values map to 0.
Now, if we pick 65534 bigrams, then there are just two remaining bigrams (due to folding) which get a priority of 0 now 🤷 .
It's mostly relelvant if we want to pick less bigrams (like 200) and even then it doesn't matter too much whether the 0 is included or not I guess.

crates/sparse-ngrams/README.md
aneubeck: I kept it as a short introduction. added two links (to our own blog post and to cursors reimplementation)

crates/sparse-ngrams/src/ngram.rs
aneubeck: added the actual reason to the comment...

aneubeck:
> This is somewhat different from what we do in Blackbird. Without getting into too much detail, we generate more n-grams.

I guess you mean being more strict when the same bigram priority occurs multiple times?
I think I did it differently back then, since the algorithm was slightly simpler. But in retrospective I don't like that this way a lot of "strange" n-grams make it into the index.

> Generally, there's an engineering tradeoff around how many to collect. More is good because queries produce fewer and rarer n-grams; but bad because indexing is more expensive and indexes are larger. It would be nice to let users choose their preferred tradeoff point, i.e. choose their desired ratio of output n-grams per byte of content. I don't know an elegant design for this, though. What we're doing in Blackbird isn't exactly elegant :)

Correct. We might want to follow up with a config object similarly to how we did it with geofilters.

PR #122, opened 2026-06-05
A reasonably fast casefold implementation

aneubeck:
This is useful in combination with sparse-ngrams indexing.
It is actually now usually 2x faster than simd-normalizer and for pure ascii even 30x faster.
This implementation is also way faster than our blackbird version.
While on a XEON processor, the throughput is lower, this implementation still wins by usually at least 2x.


crates/casefold/benchmarks/conversion.rs
```diff
@@ -0,0 +1,218 @@
...
+fn to_ascii_lowercase_into_bytes(s: String) -> Vec<u8> {
+    s.to_ascii_lowercase().into_bytes()
+}
+
+/// `simd-normalizer::casefold` — Unicode simple case folding (C+S, Unicode
+/// 17.0). The crate uses SIMD for some passes (e.g. ASCII/scan), but its
+/// case-fold mapping itself does not appear to use SIMD lookups; it is included
+/// here as another correct case-folder for throughput comparison.
+#[inline]
+fn simd_normalizer_casefold_into_bytes(s: String) -> Vec<u8> {
```
gorzell, line 51: **Nit:** The number and type of benchmarks seem to be out of sync with the comment at the top of the file.

crates/casefold/README.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+## Why does this crate exist?
+
+Unicode 16.0 defines 1484 simple-fold mappings. Common ways to store them:
+
+| Representation                                        | Size        |
+|-------------------------------------------------------|-------------|
+| Naïve `[(u32, u32); 1484]`                            | ~11.6 KB    |
+| `regex-syntax::unicode_tables::case_folding_simple`   | ~67 KB src  |
```
gorzell: **Question:** Is `src` here supposed to link to something?

crates/casefold/src/lib.rs
```diff
@@ -0,0 +1,540 @@
...
+        // plus the 4-byte store stays in bounds for every iteration. `src` and
+        // `dst` are distinct allocations.
+        unsafe {
+            let run = read - flushed;
+            if run != 0 {
+                core::ptr::copy_nonoverlapping(src.add(flushed), dst, run);
+                dst = dst.add(run);
+            }
+            // Store a full 4-byte word, advance only by the real folded length.
+            dst.cast::<u32>().write_unaligned(folded.to_le());
```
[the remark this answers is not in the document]
gorzell: There is a null check on 227.

gorzell (approved):

PR #124, opened 2026-06-10
index folding

aneubeck:
Port our index fold operation as well using the same speedup tricks.

### DIFF
--- a/crates/casefold/src/lib.rs
+++ b/crates/casefold/src/lib.rs
@@ -72,7 +72,7 @@ mod table {
 
 mod index_fold;
 mod simple_fold;
-pub use index_fold::index_fold;
+pub use index_fold::{index_fold, index_fold_char};
 pub use simple_fold::simple_fold;
 
 use table::*;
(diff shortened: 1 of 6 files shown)


crates/casefold/src/lib.rs
```diff
@@ -70,201 +70,12 @@ mod table {
...
-    // lower-case ASCII, so skipping it is purely an optimization.
-    let first_non_ascii = bytes
-        .iter()
-        .position(|&b| b & 0x80 != 0)
-        .expect("a non-ASCII byte exists (the high-bit accumulator was set)");
-    fold_non_ascii_tail(bytes, first_non_ascii)
-}
+mod index_fold;
+mod simple_fold;
+pub use index_fold::{index_fold, index_fold_char};
```
gorzell, line 75: **Question:** Why does `index_fold_char` need to be exported when the simple fold version isn't?

gorzell (approved):

crates/casefold/src/lib.rs
aneubeck: adding one for the simple fold case...
We mostly need it for the regex conversion code.

PR #135, opened 2026-07-06
Introduce a fast masked key for sorting

aneubeck:
see readme for more information.
This should speed up MST construction significantly

### DIFF
--- a/crates/geo_filters/README.md
+++ b/crates/geo_filters/README.md
@@ -61,6 +61,51 @@ Our implementation is at least as fast as the HLL++ port for all sizes.
 The `GeoDistinctCount` solution only needs to count the number of occupied buckets and the offset of the first bucket.
 Both numbers are tracked during incremental updates, such that the distinct count can be computed in constant time.
 
+## Sorting filters by masked similarity
+
+`GeoDiffCount::cmp_masked` compares two filters after applying a repeated bit mask, a locality-sensitive projection that induces a total order in which similar filters end up close together.
+Sorting or bucketing a large collection this way is dominated by the cost of the pairwise `cmp_masked` calls.
+
+`GeoDiffCount::masked_sort_key` builds a compact `u64` sort key from the most significant masked bits of a filter.
+Comparing two keys numerically reproduces the `cmp_masked` order whenever the keys differ, so a sort only falls back to the (much slower) `cmp_masked` on the rare key ties:
+
+```rust
+use geo_filters::diff_count::GeoDiffCount7;
+use geo_filters::Count;
+
+// A repeated 20-bit mask with 3 bits set acts as a locality-sensitive projection.
+const MASK: u64 = 0b0000_0100_0000_1000_0001;
+const MASK_SIZE: usize = 20;
+
+let filters: Vec<_> = (0..5u64)
+    .map(|s| {
+        let mut f = GeoDiffCount7::default();
+        (0..1000).for_each(|i| f.push(s * 1000 + i));
+        f
+    })
+    .collect();
+
+// Precompute one key per filter, then sort by comparing keys, only using the exact
+// `cmp_masked` to break ties.
+let keys: Vec<u64> = filters.iter().map(|f| f.masked_sort_key(MASK, MASK_SIZE)).collect();
+let mut order: Vec<usize> = (0..filters.len()).collect();
+order.sort_by(|&i, &j| {
+    keys[i]
+        .cmp(&keys[j])
+        .then_with(|| filters[i].cmp_masked(&filters[j], MASK, MASK_SIZE))
+});
+assert_eq!(order.len(), filters.len());
+```
+
+Sorting 1,000 filters of 1,000 items each with the 20-bit mask above, measured with `cargo bench --bench masked_sort`:
+
+| operation | `GeoDiffConfig7` | `GeoDiffConfig13` |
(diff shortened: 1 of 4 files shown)


crates/geo_filters/README.md
```diff
@@ -61,6 +61,51 @@ Our implementation is at least as fast as the HLL++ port for all sizes.
...
+
+`GeoDiffCount::masked_sort_key` builds a compact `u64` sort key from the most significant masked bits of a filter.
+Comparing two keys numerically reproduces the `cmp_masked` order whenever the keys differ, so a sort only falls back to the (much slower) `cmp_masked` on the rare key ties:
+
+```rust
+use geo_filters::diff_count::GeoDiffCount7;
+use geo_filters::Count;
+
+// A repeated 20-bit mask with 3 bits set acts as a locality-sensitive projection.
+const MASK: u64 = 0b0000_0100_0000_1000_0001;
```
itsibitzi, line 77: **Suggestion**: We could possibly have the code for generating these masks in this crate.

crates/geo_filters/README.md
```diff
@@ -61,6 +61,51 @@ Our implementation is at least as fast as the HLL++ port for all sizes.
...
+// Precompute one key per filter, then sort by comparing keys, only using the exact
+// `cmp_masked` to break ties.
+let keys: Vec<u64> = filters.iter().map(|f| f.masked_sort_key(MASK, MASK_SIZE)).collect();
+let mut order: Vec<usize> = (0..filters.len()).collect();
+order.sort_by(|&i, &j| {
+    keys[i]
+        .cmp(&keys[j])
+        .then_with(|| filters[i].cmp_masked(&filters[j], MASK, MASK_SIZE))
+});
+assert_eq!(order.len(), filters.len());
```
itsibitzi, line 97: We could have an extension trait that does this for people. It's quite verbose to ask people to do this.

crates/geo_filters/src/diff_count.rs
```diff
@@ -145,6 +145,49 @@ impl<'a, C: GeoConfig<Diff>> GeoDiffCount<'a, C> {
...
+        // The most significant bits are stored sparsely and sorted from largest to smallest, so we
+        // can test each of them against the periodic mask directly, avoiding the more expensive
+        // merge performed by `bit_chunks`. The least significant bits are dense and always below
+        // every most significant bit, so we mask whole 64-bit blocks at once. Chaining the two
+        // therefore yields the masked one-bit positions from largest to smallest.
+        let msb = self
+            .msb
+            .iter()
+            .map(|&p| p.into_usize())
+            .filter(move |&pos| (mask >> (pos % mask_size)) & 1 != 0)
```
itsibitzi, line 182: This will panic if `mask_size` is zero. The assertion Copilot recommends is probably good idea.

crates/geo_filters/README.md
itsibitzi: Copilot came up with the following:

```rust
pub trait MaskedSort<'a, C: GeoConfig<Diff>> {
    /// Returns the indices of the filters ordered by masked similarity, leaving the slice
    /// untouched. The returned permutation `order` is such that `order[k]` is the index of the
    /// filter that comes `k`-th in masked-sorted order.
    fn masked_sort_order(&self, mask: u64, mask_size: usize) -> Vec<usize>;

    /// Sorts the filters in place by masked similarity.
    fn sort_by_masked_similarity(&mut self, mask: u64, mask_size: usize);
}

impl<'a, C: GeoConfig<Diff>> MaskedSort<'a, C> for [GeoDiffCount<'a, C>] {
    fn masked_sort_order(&self, mask: u64, mask_size: usize) -> Vec<usize> {
        let keys: Vec<u64> = self
            .iter()
            .map(|f| f.masked_sort_key(mask, mask_size))
            .collect();
        let mut order: Vec<usize> = (0..self.len()).collect();
        order.sort_by(|&i, &j| {
            keys[i]
                .cmp(&keys[j])
                .then_with(|| self[i].cmp_masked(&self[j], mask, mask_size))
        });
        order
    }

    fn sort_by_masked_similarity(&mut self, mask: u64, mask_size: usize) {
        let order = self.masked_sort_order(mask, mask_size);
        // Turn `order` (source index for each destination) into the inverse permutation `pos`
        // (destination index for each source), then apply it in place with cycle swaps. This
        // reorders the filters without cloning any of them.
        let mut pos = vec![0usize; order.len()];
        for (dst, &src) in order.iter().enumerate() {
            pos[src] = dst;
        }
        for i in 0..pos.len() {
            while pos[i] != i {
                let j = pos[i];
                self.swap(i, j);
                pos.swap(i, j);
            }
        }
    }
}
```

itsibitzi: That said, it might be nice to memoize the sort key for our use case so we wouldn't use the extension trait anyways.

itsibitzi (approved):
Very neat.

crates/geo_filters/README.md
aneubeck: yeah. should be a subsequent step.
Essentially, we should have the MST computation moved in here...

crates/geo_filters/README.md
aneubeck: yeah, kinda tricky these extension traits, since they are often slightly off from what you want.
we also want a par iter version of it 🤷

PR #136, opened 2026-07-07
Closed form approximations

aneubeck:
This PR introduces a closed form approximation of the size estimation.
For the inverse, it was simpler to use Newton than trying to find another approximating function.
Note: these functions are used instead of the LUT table, since the LUT only works for a certain number of one bits.

The PR also adds documentation how the error rate increases for the metric when the set size goes up and a guideline when to use it.


itsibitzi (approved):
Let's give it a go.

PR #138, opened 2026-07-14
Latest iteration on ngrams

aneubeck:
Note: I might test a different table with more bigrams from a fresh index...
But that requires integration with the reindex code and comparing actually disk size.


crates/sparse-ngrams/src/table.rs
```diff
@@ -1,41 +1,115 @@
...
+];
+
+/// Reconstructs the priority of the ascii bigram `(a, b)`; see [`bigram_priority`]. This rolling
+/// variant avoids re-loading `BIGRAM_H[a]`: consecutive bigrams overlap by one byte, so the caller
+/// passes the `h_b` returned for the previous position as `h_a` and gets `BIGRAM_H[b]` back for the
+/// next one. The `H` value is only used for ascii bytes; for a non-ascii byte the bigram is absent,
+/// so the masked lookup (`b & 0x7f`) merely returns a value that the next step discards.
+#[inline]
+pub(crate) fn bigram_priority_rolling(a: u8, b: u8, h_a: u32) -> (u32, u32) {
+    let h_b = BIGRAM_H[(b & (BIGRAM_ALPHABET as u8 - 1)) as usize] as u32;
```
itsibitzi, line 54: Duplication of `bigram_h` below.

crates/sparse-ngrams/README.md
```diff
@@ -31,39 +33,42 @@ let input = b"hello world";
...
-| 400       | 6.8M            | 90.2%    |
-| 800       | 7.3M            | 96.0%    |
-| 1,600     | 7.5M            | 99.2%    |
-| 3,200     | 7.6M            | 99.9%    |
-| 5,845     | 7.6M            | 100%     |
+## Bigram priority model
 
-The current bigram table contains the 5,845 most frequent bigrams from a large code corpus.
-The table saturates quickly — the first ~1,600 bigrams already capture 99% of the unique n-grams.
+Priorities come from a compact factored model (~8.5 KB) rather than a full 256×256 lookup table (~64 KB in memory). The ASCII bigram `(a, b)` is scored as `H[a] + H[b] + (code << 10) + 1`, where `H` is a shared 128-entry per-byte weight and `code` is a 4-bit per-bigram correction; a per-bigram index folded into the low bits makes every priority unique while a higher score still means a more frequent bigram. The model was trained offline against a frequency ranking from a large code corpus (~1.4% inversions vs. the exact ranking).
```
itsibitzi, line 71: Should this be committed?

crates/sparse-ngrams/src/table.rs
```diff
@@ -1,41 +1,115 @@
...
+    // keeps every priority unique.
+    let base = h_a + h_b + ((code as u32) << BIGRAM_CODE_SHIFT) + 1;
+    ((base << 16) | idx as u32, h_b)
+}
+
+/// The `BIGRAM_H` weight of a single byte, used to seed [`bigram_priority_rolling`].
+#[inline]
+pub(crate) fn bigram_h(a: u8) -> u32 {
+    BIGRAM_H[(a & (BIGRAM_ALPHABET as u8 - 1)) as usize] as u32
+}
```
itsibitzi: **Mega nit**: We use `a` everywhere to refer to the first element of the bigram, but I think this function ought to be used for both a and b. Maybe we can rename the variable here to just `byte` or something?

itsibitzi (approved):

PR #139, opened 2026-07-17
implement extraction of query ngrams

aneubeck:
This is the query part which needs to be integrated into blackbird next.

### DIFF
--- a/crates/sparse-ngrams/README.md
+++ b/crates/sparse-ngrams/README.md
@@ -46,7 +46,7 @@ invoked once per n-gram in emission order:
 use sparse_ngrams::{collect_sparse_grams_deque, NGram};
 
 let mut count = 0;
-collect_sparse_grams_deque(b"hello world", |gram: NGram| {
+collect_sparse_grams_deque(b"hello world", |gram: NGram, _idx| {
     count += 1;
     // ... insert `gram` into an index, hash it, etc.
 });
--- a/crates/sparse-ngrams/README.md
+++ b/crates/sparse-ngrams/README.md
@@ -53,6 +53,42 @@ collect_sparse_grams_deque(b"hello world", |gram: NGram, _idx| {
 assert!(count > 0);
 ```
 
+### Query-time extraction
+
+`collect_sparse_grams` emits *every* candidate gram, which is what you want when building an index.
+At query time you instead want the *minimum* set of grams that still covers the query string, fed
+incrementally as the user types. `QueryGrams` is a streaming state machine for exactly that: it
+accepts one character (or already index-folded byte) at a time, emits grams as soon as they are
+determined, and can be `flush`ed to drain the tail.
+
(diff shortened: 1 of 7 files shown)


crates/sparse-ngrams/src/query.rs
```diff
@@ -0,0 +1,656 @@
...
+    }
+
+    fn front_value(&self) -> u32 {
+        // Callers only read the front priority when the queue is non-empty; the empty case is
+        // handled earlier and must never reach here.
+        debug_assert!(!self.is_empty());
+        self.val_buf[self.head as usize]
+    }
+
+    fn pop_front(&mut self) -> u32 {
```
itsibitzi: **Nit**: Should this be `pop_front_idx`? It returns an index, not a value, and the other functions in this type are named to clarify.

crates/sparse-ngrams/src/query.rs
```diff
@@ -0,0 +1,656 @@
...
+    content_end_idx: u32,
+    /// Rolling `H` value of the first byte of the next bigram to score.
+    h: u32,
+}
+
+impl Eq for QueryGrams {}
+
+impl PartialEq for QueryGrams {
+    fn eq(&self, other: &Self) -> bool {
+        self.state() == other.state()
```
itsibitzi, line 113: This will return true for different `QueryGrams` if we've only pushed bigrams, is that intended?

crates/sparse-ngrams/src/query.rs
```diff
@@ -0,0 +1,656 @@
...
+            queue: Queue::new(),
+            content_end_idx: 0,
+            h: 0,
+        }
+    }
+}
+
+impl QueryGrams {
+    #[inline]
+    pub fn state(&self) -> (u32, u64) {
```
itsibitzi, line 154: **Nit**: A comment here explaining what this tuple represents would be helpful since its on the public API, users might want to know what its for.

crates/sparse-ngrams/README.md
itsibitzi, line 1: **Nit**: Maybe an example of using the query gram extractor would be nice in the readme?

crates/sparse-ngrams/src/query.rs
```diff
@@ -0,0 +1,656 @@
...
+
+impl PartialEq for QueryGrams {
+    fn eq(&self, other: &Self) -> bool {
+        self.state() == other.state()
+    }
+}
+
+impl Hash for QueryGrams {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.state().hash(state);
```
itsibitzi, line 119: Same here. Different bigrams will produce a `QueryGrams` that hashes to the same value.

itsibitzi (approved):

crates/sparse-ngrams/src/query.rs
aneubeck: good catch
👍

PR #144, opened 2026-08-05
Add min bloom filter crate

aneubeck:
## Summary

- add a split-block min Bloom filter storing eight nibbles per `u32`
- implement max insertion and min retrieval with an AArch64 NEON fast path
- size filters using the Poisson split-block occupancy model
- add Criterion benchmarks against `sbbf-rs-safe` over 3 million entries
- add an accuracy evaluation for geometrically distributed values

## False-positive sizing

For block occupancy `X ~ Poisson(lambda)`, the false-positive probability is:

```text
P(FP) = sum(x = 0..infinity) [
    exp(-lambda) * lambda^x / x! * (1 - (7 / 8)^x)^8
]
```

Numerically solving this expression gives 13.43 nibbles/entry for 1% FPP and 25.34 nibbles/entry for 0.1% FPP. Over 10 million absent queries, observed rates were 0.9941% and 0.0991%.

## Apple M4 Max benchmark results

| Target FPP | Min filter memory | Min insert | Min retrieve | SBBF insert | SBBF contains |
|---|---:|---:|---:|---:|---:|
| 1% | 20.15 MB | 2.71 ns/entry | 1.63 ns/entry | 1.52 ns/entry | 1.23 ns/entry |
| 0.1% | 38.01 MB | 5.02 ns/entry | 3.08 ns/entry | 1.57 ns/entry | 1.26 ns/entry |

An ordinary optimally sized Bloom filter requires 3.59 MB and 5.39 MB, respectively.

## Value accuracy

For 3 million inserted values drawn from geometric distributions with ratios 0.5 and 0.2, all 10 million inserted-key queries returned the exact value; no overestimation was observed.

Run with:

```console
cargo bench -p min-bloom-filter --bench performance
cargo run --release -p min-bloom-filter --example accuracy
```

### DIFF
--- a/crates/min-bloom-filter/README.md
+++ b/crates/min-bloom-filter/README.md
@@ -19,21 +19,87 @@ assert_eq!(filter.get(0x1234_5678_9abc_def0), 7);
 Values must be in `0..=15`. Like a regular Bloom filter, collisions can produce
 false positives: a key that was not inserted can retrieve a nonzero value.
 
+## False-positive probability and sizing
+
+Let `n` be the number of inserted keys, `B` the number of blocks, and
+`lambda = n / B` the average number of keys assigned to a block. Each block has
+eight words with eight nibbles per word.
+
+For a block containing exactly `x` keys, the probability that a selected nibble
+in one word is nonzero is:
+
+```text
+q(x) = 1 - (7 / 8)^x
+```
+
+A false positive requires the selected nibble in all eight words to be nonzero:
+
+```text
+P(FP | X = x) = (1 - (7 / 8)^x)^8
+```
+
+Block occupancy is approximately Poisson distributed, `X ~ Poisson(lambda)`.
+The filter therefore uses the split-block false-positive model:
+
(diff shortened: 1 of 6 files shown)


crates/min-bloom-filter/README.md
```diff
@@ -0,0 +1,42 @@
...
+|---|---:|---:|---:|
+| 1% | 14.52 MB | 2.27 ns/entry | 1.42 ns/entry |
+| 0.1% | 21.91 MB | 2.74 ns/entry | 1.58 ns/entry |
+
+For comparison, `sbbf-rs-safe` on the same hashes:
+
+| Target FPR | Insert | Contains |
+|---|---:|---:|
+| 1% | 1.50 ns/entry | 1.24 ns/entry |
+| 0.1% | 1.55 ns/entry | 1.26 ns/entry |
```
itsibitzi: We might want to explain that this is doing something else so the better performance isn't apples-to-apples.

PR #157, opened 2026-09-04
Publish crates to crates.io via trusted publishing

bb8gh:
Adds a `Publish crates` workflow so we can release from Actions without keeping a `CARGO_REGISTRY_TOKEN` secret around. It's `workflow_dispatch` only — pick a crate from the dropdown (or type a name into `crate_name_override` if a new crate isn't listed yet), and optionally dry-run first. `rust-lang/crates-io-auth-action` swaps the job's OIDC identity for a token that's revoked when the run ends.

The job runs in the `crates-io` environment, which is also part of what crates.io matches on. Without it, anyone who can push a branch could dispatch a modified workflow and publish from it.

Also marks the benchmark/test helper crates `publish = false` so they can't be published by accident, and adds a short release section to CONTRIBUTING.

Still needs a trusted publisher configured per crate at `crates.io/crates/<crate>/settings/trusted-publishing` (repo `github/rust-gems`, workflow `publish-crates.yaml`, environment `crates-io`).

### DIFF
--- a/.github/workflows/publish-crates.yaml
+++ b/.github/workflows/publish-crates.yaml
@@ -1,0 +1,92 @@
+# Publishes a single crate from this workspace to crates.io using Trusted Publishing.
+#
+# There is no `CARGO_REGISTRY_TOKEN` secret: `rust-lang/crates-io-auth-action` exchanges this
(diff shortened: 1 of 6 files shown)


look (approved):
Sweet

